PR #146, opened 2014-11-24
WIP: Add spawn class based on subprocess.Popen

takluyver:
This will ultimately allow simple uses of Pexpect on Windows. Ping @blink1073 - I copied your gist for the implementation of this.

Open questions:

- At present, `__init__.py` imports `pty_spawn`, so you can't even load Pexpect on Windows. I started doing something along the lines of:

```python
if sys.platform == 'win32':
    from .popen_spawn import PopenSpawn as spawn
else:
    from .pty_spawn import spawn
```

But that didn't feel quite right; an apparently similar API would be doing something very different internally, which would doubtless lead to surprising bugs. Maybe on Windows the top-level package simply shouldn't have `spawn` and `spawnu`, and we'll leave them in place on Unix for backwards compatibility.

- I'm aiming to make it as close as possible to the applicable parts of the pty_spawn API, but in a couple of places, the Unixy nature of that API conflicts with Popen and cross platform support:
   1. `spawn.kill()` takes a signal number. `Popen()` has `.kill()` and `.terminate()`, which send fixed Unix signals (SIGKILL and SIGTERM), and both do TerminateProcess on Windows.
   2. `spawn.wait()` returns the raw exit code, a 2-byte number where the high byte is the terminating signal number. `Popen.wait()` uses negative numbers for terminating signals.

   My inclination in these cases is that PopenSpawn should stay closer to the cross platform API of Popen, rather than trying to closely copy the pty_spawn interface and being inconsistent between platforms.

### DIFF
--- a/pexpect/popen_spawn.py
+++ b/pexpect/popen_spawn.py
@@ -1,0 +1,83 @@
+"""Spawn interface using subprocess.Popen
+"""
+import os
+import threading
+import subprocess
+import sys
+
+try:
+    from queue import Queue  # Python 3
+except ImportError:
+    from Queue import Queue  # Python 2
+
+from .spawnbase import SpawnBase
+
+class PopenSpawn(SpawnBase):
+    def __init__(self, cmd, timeout=30, maxread=2000, searchwindowsize=None,
+                 logfile=None, **kwargs):
+        super(PopenSpawn, self).__init__(timeout=timeout, maxread=maxread,
+                searchwindowsize=searchwindowsize, logfile=logfile)
+                
+        kwargs.update(dict(bufsize=0, stdin=subprocess.PIPE,
+                           stderr=subprocess.STDOUT,
+                           stdout=subprocess.PIPE))
+
+        if sys.platform == 'win32':
+            startupinfo = subprocess.STARTUPINFO()
+            startupinfo.dwFlags |= subprocess.STARTF_USESHOWWINDOW
+            kwargs['startupinfo'] = startupinfo
+            kwargs['creationflags'] = subprocess.CREATE_NEW_PROCESS_GROUP
+
+        self.proc = subprocess.Popen(cmd, **kwargs)
+        self._buf = ''
+
+        self._read_queue = Queue()
+        self._read_thread = threading.Thread(target=self._read_incoming)
+        self._read_thread.setDaemon(True)
+        self._read_thread.start()
+
+    def read_nonblocking(self, n):
+        orig = len(self._buf)
+        while 1:
+            try:
+                self._buf += self._read_queue.get_nowait()
+            except Queue.Empty:
+                return
+            else:
+                if len(self._buf) - orig >= n:
+                    return
+
+    def _read_incoming(self):
+        """Run in a thread to move output from a pipe to a queue."""
+        while 1:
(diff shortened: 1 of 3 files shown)


jquast:
I've reviewed this PR, +1

takluyver:
PopenSpawn seems to be hitting EOF prematurely on PyPy. Not sure what's causing that at the moment.

blink1073:
I have no experience with PyPy, does it handle `b''` strings?

takluyver:
I think so - I'd expect a very different error message if that was the problem. Maybe it's some kind of race condition, but I can't think what.

pexpect/popen_spawn.py
```diff
@@ -0,0 +1,162 @@
...
+        self._log(buf, 'read')
+        return buf
+
+    def _read_incoming(self):
+        """Run in a thread to move output from a pipe to a queue."""
+        fileno = self.proc.stdout.fileno()
+        while 1:
+            buf = b''
+            try:
+                buf = os.read(fileno, 1024)
```
jquast: @blink1073 why did you add the ``time.sleep(0.001)`` call after queue.put? I consider your original commit also correct: as our read call is always blocking here, there is no need to artificially prevent a "busyloop thread", or is there a consideration here?

jquast:
Re-reviewed with fine comb for cause of EOF on PyPy cause, I don't see it, unfortunately. I'd sooner add another SkipTest for PyPy.

pexpect/popen_spawn.py
```diff
@@ -0,0 +1,162 @@
...
+            except OSError as e:
+                self._log(e, 'read')
+
+            if not buf:
+                self._read_queue.put(None)
+                return
+
+            self._read_queue.put(buf)
+            time.sleep(0.001)
```
jquast: @blink1073 why did you add the ``time.sleep(0.001)`` call after queue.put? I consider your original commit also correct: as our read call is always blocking here, there is no need to artificially prevent a "busyloop thread", or is there a consideration here? [edit: cite correct line number]

blink1073: Just defensive programming as far as I remember.

takluyver: I'm going to try removing it to see if it makes a difference to the failing test - if it doesn't, I'll remove the commit.

takluyver:
Taking the sleep out appears to have made the same failure appear on CPython. Can anyone make anything of that?

blink1073:
Thoughts: `bufsize = 0` results in a busy-loop, and pypy "optimizes" the `sleep` out?

takluyver:
It turned out that it was revealing a real bug. When the reader thread reached EOF, `read_nonblocking()` could raise EOF while it had buffered data that had not yet been passed on to the caller. There were also different stages of the pipeline clobbering the `.buffer` attribute for one another.

I think I have fixed these things now, and the tests are passing even on PyPy. But it's very tricky to work on this code and keep all the different moving bits in my head (it's not spaghetti code, but lasagne code - lots of little layers). I feel like a much cleaner, less stateful API could be designed to do what's needed, but I'm not sure I want to write a successor to Pexpect.

blink1073:
Tricky, but your fix looks good to me.

Issue #132, opened 2014-10-31
should be possible to set interact(escape_character=None) to disable escaping entirely


jquast:
minor enhancement, this fails in both python3:
```
self.__interact_copy(escape_character, input_filter, output_filter)
AttributeError: 'NoneType' object has no attribute 'encode'
```

and 2:
```
-> 1695                 i = data.rfind(escape_character)
   1696                 if i != -1:
   1697                     data = data[:i]

TypeError: expected a character buffer object
```

marcosps:
Hi, @jquast 
I'm trying to fix both places and I have a patch here, but how can I test pexpect in my local repo insteado of the default pexpect installed in my system?

jquast:
using virtualenv http://docs.python-guide.org/en/latest/dev/virtualenvs/

```
pip install virtualenv
virtualenv ENV
. ENV/bin/activate
pip install ptyprocess pytest
py.test
```

marcosps:
Hi @jquast ,

currently I dont't have time do adjust the tests to be able to run interact with None as escape character. Can you please handle this?

The patch now is ok for you? Thanks a lot!
(I worked with pull requests, so if I did something wrong, please let me know.)

jquast:
Yes I will handle it soon, It's fine as-is, just do not delete your repo/branch.

Thanks again

PR #167, opened 2015-01-07
Fix problem when call interact with escape_character=None #132

marcosps:
Hi @jquast ,

thanks for the tip about virtualenv. I coudl test the change, and it seems it worked as expected. If you find something that can be better, please tell me.

### DIFF
--- a/pexpect/pty_spawn.py
+++ b/pexpect/pty_spawn.py
@@ -677,11 +677,12 @@ class spawn(SpawnBase):
         the stdout and stderr output of the child process is printed. This
         simply echos the child stdout and child stderr to the real stdout and
         it echos the real stdin to the child stdin. When the user types the
-        escape_character this method will stop. The default for
-        escape_character is ^]. This should not be confused with ASCII 27 --
-        the ESC character. ASCII 29 was chosen for historical merit because
-        this is the character used by 'telnet' as the escape character. The
-        escape_character will not be sent to the child process.
+        escape_character this method will stop. Escaping may be disableall-together
+        by setting escape_character=None. The default for escape_character
+        is ^]. This should not be confused with ASCII 27 -- the ESC character.
+        ASCII 29 was chosen for historical merit because this is the character used
+        by 'telnet' as the escape character. The escape_character will not be sent
+        to the child process.
 
         You may pass in optional input and output filter functions. These
         functions should take a string and return a string. The output_filter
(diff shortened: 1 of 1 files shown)


jquast:
Thank you for your contribution Marcos! 

I would like the following changes:
- explicitly check if ``escape_character is None``
- docstring describing that "escaping may be disabled all-together by setting ``escape_character=None``.
- use long-hand ``if`` statement instead of shorthand.

More difficult, I would like to see a test -- the file ``tests/test_interact.py`` provides one for called ``test_interact_escape``, the supporting program ``interact.py`` should be provided to evaluate whether an optional argument ``--noescape`` is used, which when it is, instead uses ``p.interact(escape_character=None)``. And a supporting test that executes this script with that argument and asserts that the default escape character is not evaluated.

I trust the code works fine, the purpose of such tests is to ensure changes or refactoring in the future does not accidentally break your feature.

You can do any or none of these, I will complete them otherwise. Thanks again!

marcosps:
Hi @jquast ,

thanks for your comments. I'll rework my patch and send it again ASAP.

marcosps:
And I'll work in all things that you pointed. Let all this to me :)

marcosps:
Ops, now it seems the unittest is getting an error. I'll try to work on those today @jquast .

BTW, I addressed your comments about pty_spawn.py, maybe some adjustments in test_interact should fix this...

pexpect/pty_spawn.py
```diff
@@ -677,11 +677,12 @@ def interact(self, escape_character=chr(29),
         the stdout and stderr output of the child process is printed. This
         simply echos the child stdout and child stderr to the real stdout and
         it echos the real stdin to the child stdin. When the user types the
-        escape_character this method will stop. The default for
-        escape_character is ^]. This should not be confused with ASCII 27 --
-        the ESC character. ASCII 29 was chosen for historical merit because
-        this is the character used by 'telnet' as the escape character. The
-        escape_character will not be sent to the child process.
+        escape_character this method will stop. Escaping may be disableall-together
```
jquast: I realize my quick draft resulted in this, but please: ``"disableall-together" => "disabled entirely"``

jquast:
It looks like you just got lucky about the failing test -- it was an intermittent failure, I re-ran it and it was fine.

(As travis has lost the history, it was in the interact spawn test -- expecting an EOF but not receiving one.  The last buffer however showed something like a final "<in> end of file\r\nno coverage will be collected." sort of text that indicates that it did, in fact, exit cleanly.

Also, noticed github was not sending push notifications to teamcity for pull requests, manually added a polling VCS trigger for pull requests. It is building the SunOS/OSX/Linux matrix now.)

marcosps:
Hi @jquast ,

sorry about the comment. It was a blind change. Fixed now. So, you think the change is ready to be pushed? I can send the changes in tests in another PRs.

jquast:
sorry for not replying sooner, but it absolutely requires tests, i meant to write them but i've been hung up

Issue #169, opened 2015-01-15
pxssh should have a unicode variant, like pexpect.pxsshu


wsong:
In Python 3, if I do:
```python
s = pxssh.pxssh()
s.login(...)
s.sendline('command')
s.prompt()
assert s.before.startswith('foobar'):
```
I get ``TypeError: startswith first arg must be bytes or a tuple of bytes, not str``

This appears to be because pxssh.pxssh inherits from spawn, not spawnu.  Is there a Unicode version of pxssh available?

jquast:
There is not. This is similar to #92.

Issue #159, opened 2014-12-19
pexpect.spawn(logfile=sys.stdout) fails on python3.4, clarify documentation to use sys.stdout.buffer.


rich-pixley:
File "/usr/lib/python3.4/site-packages/pexpect/__init__.py", line 904, in _log
    self.logfile.write(s)
TypeError: must be str, not bytes

This contradicts the doc.

jquast:
This is why modules such as "six" or the py2to3 tool exist, and I highly encourage you to use it. The need for using str('text') in both python2 and 3, and being bytes-only for the former and unicode-only for the latter, and wishing your upstream API's to do the same is a path that will only lead to UnicodeDecodeError.

Nodd:
Why not add a ```encoding``` argument to ```spawn``` ? By default it would be set to ```None```, which would mean no encoding is done, so that it accepts and returns bytes. If it is set to a string like ```"ascii"``` or ```"utf-8"```, then conversions are done exactly like in ```spawnu```. This would remove the need to separate the classes and improve discovery of this option (```spawnu``` is buried deep down in the documentation). ```errors``` could be added too with no harm.
```spawnu``` would become an alias for ```spawn``` for compatibility.

takluyver:
That's actually not a bad idea - represent bytes/unicode as states within the class rather than separate classes. I have found the class hierarchy in Pexpect somewhat awkward, since for each kind of thing, there ought to be a bytes class and a corresponding unicode class (see e.g. PR #172). It would also be roughly similar to the behaviour of Py3 `open()`, which can open files either in bytes mode or text mode. On the other hand, `open()` is a function, and returns different classes based on which mode you specify, whereas `spawn` is a class itself.

@jquast , I'd welcome a second opinion.

jquast:
agreed, i like the ``encoding`` argument proposal rather than the class taxonomies

takluyver:
Great. I will aim to get round to that soon.

rich-pixley:
That doesn't help me nor the problem I'm trying to report.  While it might make for a cleaner architecture, it doesn't seem to provide any functionality that doesn't already exist.  I'd argue that it's arbitrary make-work and your time would be better spent elsewhere.

I don't mind that there are bytes and unicode interfaces.  That's all fine and lovely and in some distant future I might actually want that functionality.  But the functionality that I want *today*, and don't have, is the ability to write naively for strings.

IARI:
[...]

PR #172, opened 2015-01-19
On hold: Creates a pxsshu variant for using Unicode strings in Python 3

wsong:


jquast:
looks good -- @wsong I presume you tested this with some utf-8 or big-5 encoded server that, with pxssh, would raise UnicodeDecodeError?

What do you think @takluyver ? ok to merge by me (and add to changelog)

wsong:
@jquast Actually, I wrote this because in Python 3, child.sendline('foobar') would error because 'foobar' is a string and not a bytes object.  See https://github.com/pexpect/pexpect/issues/169.

jquast:
Yes, that is much related, and this would indeed fix it.

(note: tests.test_interact.InteractTestCase.test_interact_spawn_eof intermittently fails for different python versions on travis)

takluyver:
Fine by me. Please do add a very brief [test](https://github.com/pexpect/pexpect/blob/master/tests/test_pxssh.py) and [doc mention](https://raw.githubusercontent.com/pexpect/pexpect/master/doc/api/pxssh.rst) for the new class.

wsong:
I've added some docs and a test, but the README for running the test seems to be out of date; there is no testall.py tool.

Additionally, the "make docs" command fails with:

make: *** No rule to make target `doc/index.template.html', needed by `docs'.  Stop.

takluyver:
Thanks. You can run the tests with `py.test` and build the docs with `cd doc && make html`. I'll update the README and Makefile.

wsong:
I noticed that the test I added fails with this error in Python 2.7:

```
======================================== FAILURES ========================================
________________________ PxsshTestCase.test_fake_ssh_with_unicode ________________________

self = <tests.test_pxssh.PxsshTestCase testMethod=test_fake_ssh_with_unicode>

    def test_fake_ssh_with_unicode(self):
        ssh = pxssh.pxsshu()
>       ssh.login('server', 'me', password='s3cret')

test_pxssh.py:54: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../pexpect/pxssh.py:214: in login
    i = self.expect(["(?i)are you sure you want to continue connecting", original_prompt, "(?i)(?:password)|(?:passphrase for key)", "(?i)permission denied", "(?i)terminal type", TIMEOUT, "(?i)connection closed by remote host"], timeout=login_timeout)
../pexpect/spawnbase.py:277: in expect
    compiled_pattern_list = self.compile_pattern_list(pattern)
../pexpect/spawnbase.py:187: in compile_pattern_list
    self._pattern_type_err(p)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pexpect.pxssh.pxsshu object at 0x107c9de50>
pattern = '(?i)are you sure you want to continue connecting'

    def _pattern_type_err(self, pattern):
        raise TypeError('got {badtype} ({badobj!r}) as pattern, must be one'
                        ' of: {goodtypes}, pexpect.EOF, pexpect.TIMEOUT'\
                        .format(badtype=type(pattern),
                                badobj=pattern,
                                goodtypes=', '.join([str(ast)\
>                                   for ast in self.allowed_string_types])
                                )
                        )
E       TypeError: got <type 'str'> ('(?i)are you sure you want to continue connecting') as pattern, must be one of: <type 'unicode'>, pexpect.EOF, pexpect.TIMEOUT
```

Is there an established way in the pexpect codebase to get around this?

takluyver:
One of the string literals - probably the password - needs to be a unicode literal, i.e. `u'secret'`. This is valid on Python 2 and Python 3.3 and above, which covers all the versions we support, so it's safe to use.

wsong:
I'm still getting the same error; it seems that the .expect() calls in login don't work with str-type strings in Python 2.  Is this expected?  Why does SpawnBaseUnicode not have allowed_string_types = (str, unicode) instead of just (unicode,)?

jquast:
> Why does SpawnBaseUnicode not have allowed_string_types = (str, unicode) instead of just (unicode,)?

The base pexpect.pxssh class needs to use like:
```python
self.expect([self.string_type("(?i)are you sure you want to continue connecting"),
             ...
```

which should be bytes or unicode depending on pxssh/pxsshu (and str/unicode/bytes depending on python version)

takluyver:
That's expected, SpawnBaseUnicode is a strict unicode interface, so you need to expect unicode strings for it.

wsong:
Unfortunately, that won't work in Python 3 because string_type is bytes, and bytes(<a string>) is invalid because you need to specify an encoding (e.g. "bytes('foobar', 'utf-8')").

wsong:
So do we have an established way of building strings/bytes/unicode/whatever with a SpawnBase object?  Could we just provide a build_string() method or something that takes in a string and returns a str, unicode, or bytes object (in the latter case using the 'utf-8' encoding)?

takluyver:
We're considering a different way to distinguish bytes vs unicode in issue #159.

Issue #176, opened 2015-02-10
Allowed values for events argument in pexpect.run()


bancal:
In `pexpect/__init__.py` line 209, events's value is checked to be an instance of `types.FunctionType`.

In some case it could be usefull to give an instance's method `types.MethodType`. Are there reasons for not allowing it?

Actual behavior is that If we give a `types.MethodType`, we get the error "TypeError: The callback must be a string or function."

jquast:
probably True, could stand to be ``is callable()``.  a PR that includes tests for method would be most appreciated if you're interested.

bancal:
Just a question ... I'm not (yet) used to github practice. What do you mean by "needs-changelog"?

jquast:
need to add to doc/history.rst before merge

jquast:
I had to make another PR for the minor additions I made, Closing in preference to PR #178.

Thank you again for your contribution, this will be in pexpect release version 4.0.

PR #177, opened 2015-02-12
Allows also method callback for events argument in pexpect.run()

bancal:
As discussed here https://github.com/pexpect/pexpect/issues/176
I didn't opt for `callable()` since it can include other cases (not tested here).

I updated test_run.py for this feature.


jquast:
Closing in preference to PR #178.

Thank you very much for your contribution, @bancal. This will be in pexpect release version 4.0.

Issue #174, opened 2015-01-21
regression: fdspawn lost its send() method in 3.4


jquast:
looks like ``fdspawn`` inherits directly from ``SpawnBase``, which, although provides ``sendline``, does not provide ``send`` -- this method is only available from ``spawn``.  Seems like some oversight.

I think this is probably related a bit to issue #92, we should have, I presume, a ``fdspawn`` and ``fdspawnu`` variant, inheriting from ``spawn`` and ``spawnu``, respectively -- not ``SpawnBase``.

```
Date: Wed, 21 Jan 2015 10:35:10 -0500
From: Aaron <xxx>
To: contact@jeffquast.com
Subject: fdspawn in pexpect 3

Hi there,

I'm converting a project that used pexpect 2.4 to pexpect 3.2 and I'm hung
up on changes made to fdspawn. It seems that fdspawn no longer has the
'send' method:

  File "/Users/XXX/YYY/cpelib.py", line 6, in send_uboot_cmd
    exp.send(cmd)
AttributeError: 'fdspawn' object has no attribute 'send'

Is this a known problem or have I done something wrong? Looking at the
code, SpawnBase doesn't have send(),  this is added one level up in spawn,
so the error seems reasonable.

Thanks for any help you can provide.

cheers,
Aaron
```

takluyver:
I think that was fairly deliberate - fdspawn should only assume that the fd it's given is readable, not that it's also writeable. And if you need to write to it, you can use `os.write()` - there's no need to involve pexpect in matters. I'm comfortable with breaking a few cases like this for Pexpect 4.

That shouldn't have affected 3.2, though - the refactoring that created SpawnBase is more recent than that.

jquast:
It was available in pexpect 2.4 per the e-mail

jquast:
Also, why does ``sendline`` work, then?

takluyver:
Does it? I don't see a `sendline` method on either SpawnBase or fdspawn in master.

In 3.2 it may simply have been an oversight that send was removed.

jquast:
version 3.2 on pypi has both ``send`` and ``sendline`` https://github.com/pexpect/pexpect/issues/161#issuecomment-67595218

[...]

PR #178, opened 2015-02-15
Support MethodType as callback for pexpect.run(event=...)

jquast:
For your quick review, @takluyver: This supersedes PR #177 and closes issue #176.

**major**
- bumps ``__version__`` to 4.0.dev as discussed in https://github.com/pexpect/pexpect/issues/174
- allows a method to be used as an event callback for ``run()``

**minor**
- better TypeError for invalid callback types
- more coverage to include this TypeError, ``test_run_event_typeerror``
- enhance readability of ``test_run.py``

### DIFF
--- a/doc/history.rst
+++ b/doc/history.rst
@@ -12,6 +12,8 @@ Version 4.0
   coroutine. You can get the result using ``yield from``, or wrap it in an
   :class:`asyncio.Task`. This allows the event loop to do other things while
   waiting for output that matches a pattern.
+* Enhancement: allow method as callbacks of argument ``events`` for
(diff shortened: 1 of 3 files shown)


takluyver:
Looks good to me. Just prodded a failing Travis test job to re-run - it's probably just a sporadic race condition - so long as it passes, I think this is fine to merge.

We may at some point want to expand the definition of callbacks to just check for a callable rather than specific types, but if we've got this far with just functions, there's probably no rush.

jquast:
I'm thinking about making this "intermittently failing EOF test" allow either "<eof>" string (currently) or simply an actual EOF. PR soon.

PR #182, opened 2015-02-27
'Unicode mode' in spawn class

takluyver:
As discussed, this makes unicode vs. bytes two modes within spawn, rather than separate classes. There are various spawn classes for other purposes, and we were heading towards having fiddly multiple inheritance to get pxsshu, fdspawnu, etc.

I've left `spawnu` and `runu` as wrapper functions for backwards compatibility, but marked them deprecated and removed them from docs. They could alternatively be left as convenience functions: `spawnu(...)` is a bit less typing than `spawn(..., encoding='utf-8')`. I don't have a strong preference on this, but I've left updating the tests until we make a decision on this.

Also still to do: expose the encoding & codec_errors parameters on pxssh, fdspawn, and anything else that may need them.

Testing this requires the corresponding branch of ptyprocess (pexpect/ptyprocess#10).

### DIFF
--- a/doc/api/pexpect.rst
+++ b/doc/api/pexpect.rst
@@ -36,9 +36,9 @@ spawn class
 
       .. note::
 
-         With a :class:`spawn` instance, the log files should be open for
-         writing binary data. With a :class:`spawnu` instance, they should
-         be open for writing unicode text.
+         With :class:`spawn` in bytes mode, the log files should be open for
+         writing binary data. In unicode mode, they should
+         be open for writing unicode text. See :ref:`unicode`.
 
 Controlling the child process
 `````````````````````````````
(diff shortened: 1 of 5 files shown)


doc/api/pexpect.rst
```diff
@@ -69,31 +69,35 @@ Controlling the child process
...
 
-For backwards compatibility, :class:`spawn` can handle some Unicode: its
-send methods will encode arbitrary unicode as UTF-8 before sending it to the
-child process, and its expect methods can accept ascii-only unicode strings.
-However, for a proper unicode API to a subprocess, use this subclass:
+By default, :class:`spawn` is a bytes interface: its read methods return bytes,
+and its write/send and expect methods expect bytes. If you pass the *encoding*
+parameter to the constructor, it will instead act as a unicode interface:
+strings you send will be encoded using that encoding, and bytes received will
+be decoded before returning them to you. In this mode, patterns for
```
jquast: good description

jquast:
I'll try to figure how to make the CI's figure out how to test the two together, and see where the coverage falls in regards to new + legacy auto-encoding. Would probably work better if ptyprocess were managed as a git submodule, so that they could be tracked together.

takluyver:
Actually, thinking about it some more, these changes are not coupled to pexpect/ptyprocess#10 after all. The code in Pexpect implements encoding and decoding separately from ptyprocess, and always uses the bytes interface of PtyProcess.

takluyver:
It's now passing on Travis, though I had to rerun a couple of the builds to get that.

takluyver:
We've agreed that this is the best approach, so I'm merging this and we can push forwards on top of it.

PR #193, opened 2015-03-27
Fix syntax highlighting for LICENSE in README

nmadhok:
Currently it looks like this:

screen shot 2015-03-27 at 4 00 14 am


jquast:
Thanks!

PR #201, opened 2015-04-05
Set `flag_eof` in expect.py when raising EOF. Fixes #200.

### LINKED ISSUE #200: p.eof() returns False after pexpect.EOF raised (async use)
The [`spawn.eof()` docs](https://pexpect.readthedocs.org/en/latest/api/pexpect.html#pexpect.spawn.eof) say:

> This returns `True` if the EOF exception was ever raised.

I created a [simple script](https://gist.github.com/detly/8f936259753e4a23b50c) that seems to break this when `async=True` is used: the EOF exception is caught, but it looks like `p.flag_eof` is still `False`.

The only place I could find in the code that raises EOF *without* first setting `flag_eof = True` is in [`expect.py:49`](https://github.com/pexpect/pexpect/blob/master/pexpect/expect.py#L49). I'll see if tweaking this fixes the problem, but I'm not 100% sure it's the right solution.

detly:
Not 100% sure this is the right thing to do, but it works.

### DIFF
--- a/pexpect/expect.py
+++ b/pexpect/expect.py
@@ -46,6 +46,7 @@ class Expecter(object):
             msg = str(spawn)
             if err is not None:
                 msg = str(err) + '\n' + msg
+            spawn.flag_eof = True
             raise EOF(msg)
     
     def timeout(self, err=None):


detly:
Just saw comment on #200 — ignore this :)

PR #207, opened 2015-04-17
Fix issue 126 for 3.x branch

Depado:
It has been implemented for 2.x version but not 3.x.

### DIFF
--- a/pexpect/__init__.py
+++ b/pexpect/__init__.py
@@ -490,7 +490,10 @@ class spawn(object):
         # inherit EOF and INTR definitions from controlling process.
         try:
             from termios import VEOF, VINTR
-            fd = sys.__stdin__.fileno()
+            try:
+                fd = sys.__stdin__.fileno()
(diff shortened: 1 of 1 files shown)


takluyver:
I'm not sure if there will be another 3.x release, but I'll merge this in case there is.

@jquast we should work out where we are and try to get 4.0 out. I've just got back from Pycon, so give me a few days to try to dig myself out.

Depado:
Is there a really big gap between the 2.x and 3.x versions ? If not maybe I could work on a solution that fits both using Six for example. What do you think ?

takluyver:
The 3.x branch is for Pexpect version 3.x, not Python 3.x. It's already compatible with both Python 2 and 3. We're actually heading for version 4, which is why I said there may not be another 3.x release. I was wondering why you were talking about 2.x... ;-)

Depado:
Oh right. I didn't get it :)
Thanks for the amazing work.

PR #208, opened 2015-04-20
Escape quote in print() and fix indexes

epadillas:
Fix "bugs" in pexpect/doc/overview.rst


jquast:
Thank you @epadillas !

Issue #210, opened 2015-04-26
Build Infrastructure: Allow Travis-CI builds to comprehend ptyprocess relationship


jquast:
TeamCity comprehends the relation of pexpect and ptyprocess, and if two branches are made of the same name, they are tested together.  Travis-CI however does not, which would require publishing ptyprocess to the public to test a pairing branch in pexpect.

Travis-CI supports any kind of linux command, so we should be able to solve this with a little bit of shell scripting and a series of git(1) commands.

takluyver:
I think it's valuable to have some reminder that this doesn't work with the released version of ptyprocess. I hope that, going forwards, the projects won't be so tightly coupled that they regularly need matching changes. Of course there are some teething issues as we separate them out.

PR #211, opened 2015-04-26
Do not raise Exception on wait() after terminate or previous wait()

jquast:
This matches the same branch of ptyprocess, https://github.com/pexpect/ptyprocess/pull/18 from @reynir, ``noexception-on-wait-after-terminate``, which allows calling the wait() method multiple times without raising an exception.

There is no logical change, only a docstring and adjustments to (otherwise failing) test cases.

(@takluyver:  Note that TeamCity comprehends "pairing branch names" across these two repositories and tests them accordingly, but Travis-CI does not (so the test will fail, here). Created issue #210 to address.

Furthermore, described in https://github.com/pexpect/pexpect/commit/4bd7305df7a582887db7202128f285a2970aca6f it is now necessary to specify ``>=0.5`` requirement of ptyprocess.  In this particular case, It isn't terribly harmful, but it is good to start.  This too will break the travis-ci build, which would be resolved by release of ptyprocess 0.5 or issue #210)

### DIFF
--- a/doc/history.rst
+++ b/doc/history.rst
@@ -14,6 +14,9 @@ Version 4.0
   waiting for output that matches a pattern.
 * Enhancement: allow method as callbacks of argument ``events`` for
   :func:`pexpect.run` (:ghissue:`176`).
+* It is now possible to call :meth:`~.wait` multiple times, or after a process
+  is already determined to be terminated without raising an exception
+  (:ghpull:`211`).
 
 Version 3.4
 ```````````
--- a/pexpect/pty_spawn.py
+++ b/pexpect/pty_spawn.py
@@ -614,10 +614,17 @@ class spawn(SpawnBase):
         not read any data from the child, so this will block forever if the
         child has unread output and has terminated. In other words, the child
         may have printed output then called exit(), but, the child is
-        technically still alive until its output is read by the parent. '''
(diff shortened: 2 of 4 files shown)


takluyver:
This looks reasonable. We need to actually get round to releasing ptyprocess.

jquast:
Do we have to wait until ptyprocess is released to merge?

takluyver:
I'd prefer to keep pexpect master building against released versions of Ptyprocess. As mentioned on #210, it's sort of deliberately inflicted pain to try to ensure a clean distinction between the two packages. I don't think a ptyprocess release is far off, though - we really just need to work out pexpect/ptyprocess#12 and make a decision on pexpect/ptyprocess#10, as far as I know.

jquast:
That's fine. I'm working out https://github.com/pexpect/ptyprocess/issues/12 now.

Bumping into a strange issue with a new test case I introduced where getecho() returns False on FreeBSD and receiving echo of input in read() calls, even though PtyProcess([cmd], echo=True) is used.

I'm working it out with TeamCity to run some tests across the board for me, I hope to address it soon.

takluyver:
I reran Travis now that Ptyprocess 0.5 is released, and it is now happy with this.

Issue #215, opened 2015-05-11
performance problems due to time.sleep(0.0001) in expect_loop


billpage:
Is there a good reason to keep this:
https://github.com/pexpect/pexpect/blob/3.x/pexpect/__init__.py#L1540
Is there really an issue with "busy waiting" given the way ''read_nonblocking' is written?
See discussion at: http://trac.sagemath.org/ticket/10295#comment:50

jquast:
We need to periodically check os.waitpid for child process termination, which is done in each loop.

billpage:
Would ''time.sleep(0)'' be sufficient? 
http://stackoverflow.com/questions/7273474/behavior-of-pythons-time-sleep0-under-linux-does-it-cause-a-context-switch

jquast:
Regarding the sage math conversation, I would be a proponent of removing it entirely, if possible.

We have a large suite of tests, if they pass (a PR will trigger an OS X, Linux, FreeBSD, and Solaris build farm)., it can be removed.  If not, we need to figure out why noah felt he needed to add it and possibly document.

You would need to also install a pairing ‘master’ branch of https://github.com/pexpect/ptyprocess/ <https://github.com/pexpect/ptyprocess/> to test locally

jquast:
``sleep(0)`` may be sufficient, if it can be proven it "yields" a python thread's control on so all platforms? what a crummy hack, could use a comment above it describing why it exists!

nexttime:
> We need to periodically check os.waitpid for child process termination, which is done in each loop.

Well, that's rather orthogonal to the `sleep()` call.

Also, `isalive()` (which calls `waitpid()`) is called unconditionally each time **before** an attempt to read is made;  IMHO that should only be done after a failed read / timed-out `select()`.

jquast:
You're welcome to try any such changes, the test results of the build servers will help support or deny your suggestions as a pull request.

billpage:
This pull request
https://github.com/pexpect/pexpect/pull/216
gave mixed results. I do not understand how to interpret the ca/teamcity result.

billpage:
@jquast wrote:
"You would need to also install a pairing ‘master’ branch of https://github.com/pexpect/ptyprocess/ <https://github.com/pexpect/ptyprocess/> to test locally"

Could you explain this in a bit more detail or point me to some docs?  The README in the /tests directory seems out of date.  Something is missing from the /tools directory. Apparently I can no  longer test pexpect that way?

[...]

PR #216, opened 2015-05-12
Do we really need 'time.sleep()' in expect_loop?

billpage:
https://github.com/pexpect/pexpect/issues/215


billpage:
I do not know how to interpret the ca/teamcity result.

jquast:
Not sure yet if we'll continue cutting 3.x, if we do, we need to transfer testing framework over. For the time being, causing some issues with sigwinch in py3.4. #215 when compared against master

jquast:
Duplicate of #217

PR #217, opened 2015-05-12
Do we really need 'time.sleep()'?

billpage:
https://github.com/pexpect/pexpect/issues/215

Tried this on 3.x branch but ci testing apparently does not work on 3.x, so here we are again against master branch.

### DIFF
--- a/pexpect/expect.py
+++ b/pexpect/expect.py
@@ -95,6 +95,7 @@ class Expecter(object):
                     return self.timeout()
                 # Still have time left, so read more data
                 incoming = spawn.read_nonblocking(spawn.maxread, timeout)
+                time.sleep(0)
                 if timeout is not None:
                     timeout = end_time - time.time()
         except EOF as e:
--- a/pexpect/expect.py
+++ b/pexpect/expect.py
@@ -95,7 +95,6 @@ class Expecter(object):
                     return self.timeout()
                 # Still have time left, so read more data
                 incoming = spawn.read_nonblocking(spawn.maxread, timeout)
-                time.sleep(0.0001)
                 if timeout is not None:
                     timeout = end_time - time.time()
         except EOF as e:
--- a/pexpect/expect.py
+++ b/pexpect/expect.py
@@ -95,7 +95,7 @@ class Expecter(object):
                     return self.timeout()
                 # Still have time left, so read more data
                 incoming = spawn.read_nonblocking(spawn.maxread, timeout)
-                time.sleep(0)
+                time.sleep(0.0001)
                 if timeout is not None:
                     timeout = end_time - time.time()
         except EOF as e:


jquast:
I was just coming to see in amazement why sigwinch wasn't failing in the currently running tests, but I see it is reverted to its (0.00001) bit. Well, the scope is right, then.

billpage:
I just wanted to reassure myself that (0.00001) made all the difference so I did the last commit to get back to "square 1".  Is sigwinch the only test that uses a helper script?  Is it the only one that uses it's own timer?  It seems odd that it passes the first expect for "READY" and then fails on EOF waiting for a timer triggered result.

billpage:
I can not reproduce the failure in test_winsize.py using Python 3.3 on OpenSuse.  The test passes with the time.sleep() deleted on my local system.  Could this be a docker artifact? Or something similar?

jquast:
travis-ci might be docker, where it is passing. Is your OpenSuse a docker?

teamcity's: linux and freebsd host is kvm (digital ocean),
a paravirtualized zone for solaris (joyent)
and a vmware host for OSX (hostmyapple)

billpage:
My OpenSuse 13.1 system is running under VirtualBox on a SLES 11/sp3 server.  I do not know how to interpret the ci/teamcity results since there seem to be test failures even after I returned the source to the original coding while the Travis build passed as expected.

billpage:
Travis worked once and failed once on the same commit.  I cannot reproduce the failure locally. Can I trust it?

#650
https://travis-ci.org/pexpect/pexpect/builds/62707898
https://github.com/pexpect/pexpect/commit/ecac053daed7d7ff0f1ba790ecf9a0724c7bd768

#647
https://travis-ci.org/pexpect/pexpect/builds/62276250
https://github.com/pexpect/pexpect/commit/6dcd9afa7ad78881396c1c0a5080203ae609a450

jquast:
You might start by first authoring a test that very directly proves the conditions that the sleep calls exists for, which reliably fails when the sleep call is removed.

Then you might understand why it is there, and how to amend or address it.

jquast:
I should probably be more direct. I don't feel comfortable installing signal handlers as a maintainer who ultimately would support any unattended side-effects. Therefore, I would accept the change if the problem (sleep call exists) and the solution (removing it) can be very accurately described, and proved by tests and reason.

I think I've made the only attempt to reason about your problem so far, I am unsatisfied.

billpage:
I am not aware of any specific situation which fails when the call to sleep
is removed, although as I recall there was a failure in the test suite at
least on some platforms.  I suppose that it would be necessary to attempt
to analyse the failing case(s) and try to distill a specific test which
reliably fails.  As you might imagine it is difficult for me to find
motivation to solve problems on a platform that I do not use.

billpage:
I understand your reluctance to change something which is not documented
and for which we have no explanation.  How about allow a new option when
instantiating pexpect that would remove the call to sleep if requested but
which would default to the current situation?

jquast:
I can imagine. In fact, I don't even use pexpect at all ... :)

jquast:
This was introduced in 600463686284ea102548f5e0bf51582db051dc78: 

```
Author: noah <noah@656d521f-e311-0410-88e0-e7920216d269>
Date:   Tue May 2 23:07:42 2006 +0000

    Added time.sleep(0.0001) as suggested by Jon Cohen.
    This helps in give other threads a chance to run.
```

I've brought this in upstream as branch "superfluous-sleep". No promises but hopefully we can do something to get rid of it!

jquast:
http://jessenoller.com/blog/2009/02/01/python-threads-and-the-global-interpreter-lock has a very good discussion on why this time.sleep(0.0001) is here.  By the way, if value 0 is used, python optimizes it away. What we're after is Py_BEGIN_ALLOW_THREADS.

jquast:
Closing in preference of #291

PR #222, opened 2015-06-09
Update `split_command_line` to use `shlex.split()`.

chaorenl:


takluyver:
If this was new code, it would definitely make sense to use shlex. But it's old code, and I'm worried about introducing subtle bugs if shlex behaves differently from the code that was already in use.

chaorenl:
shlex can't be less correct than the existing implementation. If it does behave differently, it should be fix existing subtle bugs.

takluyver:
I'm not talking about what's correct - just that this has been around for years, and whatever behaviour it has, there might be code out there that will break if it suddenly changes. It's not a maintenance burden to keep it there - I don't think I've ever had to touch that code - so I don't think there's a compelling reason to clean it up.

chaorenl:
If you ever plan to support pexpect on Windows, then this will be necessary.

> there might be code out there that will break if it suddenly changes

Then they were relying on incorrect behavior and I have no sympathy for them.

takluyver:
> Then they were relying on incorrect behavior and I have no sympathy for them.

This is not a helpful attitude when maintaining stable libraries.

jquast:
concur with @takluyver, I would accept a PR to modify the docstring to suggest using shlex instead!

> If you ever plan to support pexpect on Windows, then this will be necessary.

This is not used internally, it is a "helpful utility", the merge of winexpect would not be impacted.

(By the way, the inverse of shlex.split  is not well known, pipes.quote, strange place in a strange library)

Anyway @chaorenl please do not be discouraged, this utility is 14 years and distributed in many components, it requires maintaining state of its original utility functions such as these. I doubt shlex module even existed in it's python 2.3 era. I would delete nearly half of this repository if it were a new from-scratch :)

chaorenl:
Winexpect has an [even more convoluted version](https://bitbucket.org/geertj/winpexpect/src/4848ce3592fda49211c5a19ed17e05853eadfc70/lib/winpexpect.py?at=default#cl-67) of this utility. I still disagree with favoring stability over correctness, but this is your library, so I'll respect your decisions.

PR #236, opened 2015-09-09
Restore send/write methods to fdpexpect

takluyver:
Closes gh-174

I'm adding these back for compatibility, but I don't want to document them or encourage their use - if you're working with an fd, I think it's clearer to use os.write() directly rather than Pexpect's wrappers.

I haven't added sendintr, sendeof or sendcontrol, because I don't think they really make sense for a general fd, and you can easily do the equivalent things explicitly if you need to.


jquast:
postmortem +1

PR #239, opened 2015-09-12
Encoding parameters for pxssh and fdspawn

### LINKED ISSUE #92: fdexpect should have a unicode variant, like pexpect.spawnu
fdexpect.fdspawn is based on pexpect.spawn, which accepts bytes rather than unicode.  fdexpect should provide a unicode variant, like/based on pexpect.spawnu.

takluyver:
Allows them to provide a unicode interface in the same manner as spawn.

Merging this will close #92, #169 and #172.


jquast:
+1

PR #240, opened 2015-09-12
Deprecate terminal emulator code

takluyver:
`pexpect.screen` and `pexpect.ANSI` are probably quite buggy, and there are projects out there which focus on terminal emulation, like [pyte](https://pypi.python.org/pypi/pyte), so we decided to deprecate them.


jquast:
+1, merging

PR #249, opened 2015-09-18
Use ^C test_interact_escape_None, ^D unreliable

jquast:
Test intermittently fails PyPy and Travis-CI hosts, where EOF ins incorrectly interpreted instead of captured as exception.


jquast:
Merging by --squash, need to reduce false errors by build systems for remaining PR's

Issue #175, opened 2015-01-30
timeout=None


Nodd:
The special value ```None``` for ```timeout``` is only documented in ```waitnoecho()``` (actually timeout isn't even documented at all in ```spawn```). I only discovered it by reading https://github.com/pexpect/pexpect/issues/54#issuecomment-44801435.

On the same subject, the wait docstring states that the call to this method can hang indefinitely. Why isn't the standard behavior of wait a call to ```self.expect(pexpect.EOF, timeout=None)``` (or equivalent) ? Having a potential infinite block (even documentated) when an alternative exists seems somewhat absurd.
If it's not possible, this particular call should be made as a new method, or tied to a new ```wait``` keyword, or at a minimum be explicitely shown as an example in the documentation.

jquast:
Definitely agree the need to document ``timeout=None`` as blocking indefinitely where appropriate.

If I understand your second paragraph correctly, you propose changing the default timeout of ``30`` to ``None`` everywhere. If so, I agree.  Our next release is "4.0" and we allow some API changes there, its OK with me, how about you, @takluyver ?

Nodd:
Using None as a default timeout would be good, but that was not what I meant.

I was talking about the ```wait()``` method. In the documentation it states explicitely that if can hang indefinitely if there is some output written by the program after the call to ```wait()``` In the comment linked above, an equivalent non-hanging code is proposed. My proposal is to replace the current ```wait()``` method by this alternative, something like:
```
def wait(self):
    self.expect(pexpect.EOF, timeout=None)
```
This would have the same behavior as the current ```wait()``` function, but it would not be problematic if some output is written by the program.

Now if it is not possible, I suggest to either make it a new method, or at least add it in the documentation. Even if the code is very simple, it may not be obvious to a new user (like me!).

takluyver:
I think that is reasonable - it's hard to imagine a case where someone relies on wait() blocking indefinitely, and easy to imagine getting frustrated by that behaviour. So yes, I think it's reasonable to make wait() consume and discard any output. The change for this will now need to be in ptyprocess, though.

On the other hand, I think changing default timeouts from 30s to None (wait indefinitely) is too big a change even for 4.0. That's the sort of thing people could easily be relying on, and if they are, processes will suddenly start hanging, which is a horrible failure mode.

jquast:
[...]

PR #253, opened 2015-09-20
docstring about timeout behavior

jquast:
This addresses Issue #175 directly: It is not immediately obvious that all calls will TIMEOUT at 30 seconds, and that None may block indefinitely.

regarding the meaning of the ``timeout`` keyword argument in expect method calls, as well as the spawn class initializer argument and instance attribute. 

The duplicated but brief descriptions about searchwindowsize and timeout parameters are removed from method ``expect_list()``.


pexpect/pty_spawn.py
```diff
@@ -87,6 +87,12 @@ def __init__(self, command, args=[], timeout=30, maxread=2000,
         amount of data where you want to match. The searchwindowsize does not
         affect the size of the incoming data buffer. You will still have
         access to the full buffer after expect() returns.
+        When the keyword argument ``timeout`` is specified as a number,
+        (default: *30*), then :class:`TIMEOUT` will raise after the value
```
takluyver: TIMEOUT will *be raised*

Also, let's leave a blank line above this paragraph to separate it from the previous one.

PR #254, opened 2015-09-20
FAQ redirected to pexpect.ANSI, redirect to pyte

jquast:


doc/FAQ.rst
```diff
@@ -126,11 +126,10 @@ another application.
...
-this is easy. If it does screen-oriented output then it may work, but it
-could be hard. For example, trying to scrape data from the 'top' command
-would be hard. The top command repaints the text window.
-
-I am working on an ANSI / VT100 terminal emulator that will have methods
-to get characters from an arbitrary X,Y coordinate of the virtual screen.
-It works and you can play with it (see :mod:`pexpect.ANSI`), but I have
-no working examples at this time. 
+this is easy. If a program emits many terminal sequences, from video
+attributes to screen addressing such as those typically using curses(3x), then
```
takluyver: what's the `(3x)` here? I'm used to just calling it curses.

jquast: It is the manual section. I do this convention to indicate when the phrase is discussing something with a pairing manual on your system, for example, to describe using 'cat(1)' instead of 'cat'.

takluyver: Not being a big reader of manpages, I find this more confusing than useful, especially with the x - I've seen the numbers indicating manpages a few times, but never the `x`. And is there anything else you might mean by curses?

PR #256, opened 2015-09-21
Interact tests uses getch.py, avoids trailing output assertions on travis-ci

jquast:
Getting intermittent EOF issues previously thought to be PyPy, but more likely TRAVIS related.

Also use getch.py instead of echo_w_prompt.py so we can exit using \x00 instead of ^D or ^C


jquast:
@takluyver merging without approval for green build.

Issue #235, opened 2015-09-05
Missing options keyword argument to pxssh.pxssh


frogstarr78:
The documentation mentions suppling additional ssh options to the options keyword argument to the pxssh class, however, it errs when used and doesn't appear to be present in the class constructor method.

Error when used:

```
Traceback (most recent call last):
  File "./bin/cmd.py", line 209, in <module>
    login()
  File "./bin/cmd.py", line 170, in open_session
    go(username, password, hostname, port, params)
  File "./bin/cmd.py", line 104, in open_ssh_session
    client = pxssh.pxssh(options=params)
TypeError: __init__() got an unexpected keyword argument 'options'
```

Constructor v3.3:

    88     def __init__ (self, timeout=30, maxread=2000, searchwindowsize=None,                                                                                                                                                                                        | 
    89                     logfile=None, cwd=None, env=None):                                                                                                                                                                                                          | 
    90                                                                                                                                                                                                                                                                 | 
    91         spawn.__init__(self, None, timeout=timeout, maxread=maxread, searchwindowsize=searchwindowsize, logfile=logfile, cwd=cwd, env=env)

jquast:
Sorry for the trouble, this was merged last november, https://github.com/pexpect/pexpect/pull/150 but the pexpect 4.0 release is still not yet on pypi to include it. I hope any day soon.

Issue #179, opened 2015-02-18
Pxssh broken options in current pip version


Alexandertyler:
Hi. I've been working on pxssh for vm testing. I've hit some pretty serious issues with the way pxssh deals with ssh options on the current version released. It seems like it might be fixed in the dev version but I wanted to post this here in case anyone has this issue and the workaround that I discovered for it. Let me explain the problem first:
>I attempted to pass the pxssh shell `StrictHostKeyChecking=no +UserKnownHostsFile /dev/null` 
>In the latest version of the pexpect docs, it gives a couple of different steps to do this, none of which work. I dug into the source file that pip and even the latest tar ball installs and discovered some key differences between the docs and the latest version on github. It breaks any option for users to set ssh options that will be invoked when the command is called. You can see why this would be a big deal. Apart from the code not matching the documents it prevents users from making any sort of distributable program with pxssh because if the host key doesn't match, like when using a VM with a new drive, the programs errors out on an EOF. 

The work-around:

>When I explored the source file that pip installed, I found 

```python
self.SSH_OPTS = ("-o'RSAAuthentication=no'"
                   + " -o 'PubkeyAuthentication=no'")
# Disabling host key checking, makes you vulnerable to MITM attacks.
#                + " -o 'StrictHostKeyChecking=no'"
#                + " -o 'UserKnownHostsFile /dev/null' ")
```

>At some point there was a framework in place for SSH_OPTS. We can't just set this field after initialization though. In login(), the ssh_options to be passed on are set in the chunk 

```python
ssh_options = ''
        if quiet:
            ssh_options = ssh_options + ' -q'
        if not check_local_ip:
            ssh_options = ssh_options + " -o'NoHostAuthenticationForLocalhost=yes'"
        if self.force_password:
            ssh_options = ssh_options + ' ' + self.SSH_OPTS
        if port is not None:
            ssh_options = ssh_options + ' -p %s'%(str(port))
        if ssh_key is not None:
            try:
                os.path.isfile(ssh_key)
            except:
                raise ExceptionPxssh('private ssh key does not exist')
            ssh_options = ssh_options + ' -i %s' % (ssh_key)
        cmd = "ssh %s -l %s %s" % (ssh_options, username, server)
```

[...]

PR #257, opened 2015-09-21
remove version 3.4 from history

jquast:
@takluyver Please also configure readthedocs.org to point to a github release tag (3.3 is latest, matching pypi) to prevent issues such as #235, #179 and synchronize at next release.


takluyver:
I've tweaked it so that if you visit http://pexpect.readthedocs.org, it redirects to http://pexpect.readthedocs.org/en/stable/ (rather than `.../latest/`).

PR #258, opened 2015-09-21
old proposal: Document blocking write

jquast:
request for comments, provides simple documented recommendation rather than a startDiscardingOutput method.


jquast:
Deferring, made the osx build agent pause indefinitely, no time to debug

Issue #190, opened 2015-03-18
.logfile_send and .logfile_read ignored during .interact()


cagney:
Given this code stub:

cat logfile-ignored.py
import pexpect
import sys
p = pexpect.spawn("/bin/bash")
p.logfile_send = open("/tmp/file", "w")
p.logfile_read = p.logfile_send
print("Escape is ^]")
p.interact()

/tmp/file ends up empty.  However, any non-interactive exchanges to get written.

jquast:
use ``p.logfile`` instead of ``p.logfile_send`` or ``p.logfile_read``.

The latter two are for the ``.send*`` and ``.read*`` functions (triggered by ``.expect*``). ``.interact`` only writes to ``p.logfile``. 

Would accept a PR to update the ``interact`` method's docstring to clarify.

jquast:
I will propose to use self._log(data, 'read') and 'write', rather than document the caveat that written data (keystroke input) is not logged.

jquast:
Superseded by #259

PR #259, opened 2015-09-21
interact logs by self._log(s, 'read') and 'send'

jquast:
Closes issue #190: __interact_copy implements
its own read() and write() calls, circumventing
the self._log() system, and omitting to log data
that is sent to the child pty.

Related, resolve old docstring about a setlog()
method that was removed in SpawnBase refactor.


jquast:
Note: feel this change is too minor an enhancement to warrant a changelog item.

Issue #163, opened 2014-12-19
Document suggestion of stackoverflow for support, and farm stackoverflow for new FAQ entries


jquast:

jquast:
Superseded by #261

PR #261, opened 2015-09-21
suggest many alternatives for pexpect support

jquast:
First proposed in #163, to try to help tone down the number of general support requests.

Though I appreciate receiving feedback about the trickiest conditions of scripting a subprocess through a pseudo-terminal with pexpect, that pexpect might better explain by docstring or documentation, I would more appreciate such feedback in the form of a pull request!

### DIFF
--- a/doc/FAQ.rst
+++ b/doc/FAQ.rst
@@ -3,13 +3,13 @@ FAQ
 
 **Q: Where can I get help with pexpect?  Is there a mailing list?**
 
-A: As of 2015, the best ways to get general python support is the website
-forum stackoverflow_, the python-list_ mailing list, and the `#python`_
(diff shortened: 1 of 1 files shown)


doc/FAQ.rst
```diff
@@ -1,6 +1,17 @@
 FAQ
 ===
 
+**Q: Where can I get help with pexpect?  Is there a mailing list?**
+
+A: As of 2015, the best ways to get general python support is the website
+forum stackoverflow_, the python-list_ mailing list, and the `#python`_
```
takluyver: This all seems very generic. I'd like to point to some slightly more specific things, like the [pexpect tag on SO](http://stackoverflow.com/questions/tagged/pexpect).

Issue #168, opened 2015-01-14
Pexpect sometimes leaves child processes running when an exception occurs


stevmm:
I've noticed that sometimes pexpect doesnt close child processes. Here's an example from some code (python3) I'm running:

```
handle = pexpect.spawnu("/bin/bash", ['-c', 'stty -icanon -isig -echo && nc -l 48800'], echo=False)
try: #wait 3 seconds to make sure pexpect command has spawned
    handle.expect("waiting to time out", timeout=3)
except(pexpect.TIMEOUT):
    print("passed time out")

print(cause_an_exception)
```

This will leave a hanging nc process. Yes if I break up the spawn statement I nc doesnt hang:
```
handle = pexpect.spawnu("/bin/bash", echo=False)
handle.expect(bash_prompt)
handle.sendline("stty -icanon -isig -echo && nc -l 48800")
try: #wait 3 seconds to make sure pexpect command has spawned
    handle.expect("waiting to time out", timeout=3)
except(pexpect.TIMEOUT):
    print("passed time out")

print(cause_an_exception)
```

My first guess was that if the python script closes and you didnt call .close() on a spawn object that it would leave its child processes open, but my second example shows that isnt always the case. Any thoughts on what's going on here?

Thanks

jquast:
Which OS?

stevmm:
RHEL 6

stevmm:
While I dont understand why the behavior changes when I break up the statement between two commands, I saw issue #136 and tried setting ingore_sighup=False and it worked.

takluyver:
Maybe that makes sense. When the Python process closes, it will close the master side of the pty. The purpose of SIGHUP is to indicate that the controlling terminal has been closed, so that's what the child process will get. If it's ignoring SIGHUP, it won't be killed by that. I don't know why the different code you had above makes a difference, but I wouldn't be surprised if bash initialises something differently when you use the `-c` option.

Maybe we should consider making `ignore_sighup` False by default for Pexpect 4.0. We'd left it on by default for strict backwards compatibility, but I'm willing to cause some minor breakage for 4.0, and it's probably unexpected in most cases.

jquast:
Remaining issue: consider making ignore_sighup False by default for Pexpect 4.0.

I vote yes!

takluyver:
Yeah, we can probably change it now.

PR #266, opened 2015-09-22
Change ignore_sighup default to False

takluyver:
This means child processes will no longer ignore SIGHUP by default.

Also pulled out a 'backwards incompatible changes' section of the release notes to make it clearer what has changed.

Closes gh-168


jquast:
Will Merge as #267 shortly (duplicated to resolve merge conflicts)

PR #273, opened 2015-10-06
test_unicode: don't need to format variable

ignatenkobrain:
It's just useless.


takluyver:
I'm not sure where that came from. Thanks!

PR #278, opened 2015-10-06
Remove non-ascii characters breaking easy_install for python 2.7

sverrirab:
The other solution would be to add an encoding header.

When trying to easy_install pexpect (or load the setup.py file directly with python 2.7) you will get:

SyntaxError: Non-ASCII character '\xe2' in file setup.py on line 28, but no encoding declared; see http://www.python.org/peps/pep-0263.html for details

The characters seem unecessary so I just removed them.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,8 +24,8 @@ automated software testing. Pexpect is in the spirit of Don Libes' Expect, but
 Pexpect is pure Python.
 
 The main features of Pexpect require the pty module in the Python standard
-library, which is only available on Unix-like systems. Some features—waiting
-for patterns from file descriptors or subprocesses—are also available on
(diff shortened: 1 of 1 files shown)


takluyver:
They're not unnecessary, they're punctuation. I'll add the encoding header.

Also, be aware that it's now strongly recommended to use `pip install ...` instead of easy_install. Not that it would make a difference to this.

takluyver:
Actually, thinking about it a bit more, I think using pip would avoid this issue, because of the way it runs setup.py. In any case, I've added the encoding comment and uploaded 4.0.1 to PyPI.

sverrirab:
I was surprised by the fact that pip could install it since the file is technically not a valid python file.

Thank you for the quick fix!

Issue #213, opened 2015-04-27
Tracking asyncio/EOF issue on Python 3.4.3 (and not 3.4.2-)


jquast:
I've localized the problem to Python 3.4.3 on OSX and SunOS (where 3.4.3 is available). The debian linux and Travis-CI hosts are using an earlier version of python3.4 where this problem is not present.

3.4.1
-------
```
============================================ test session starts =============================================
platform darwin -- Python 3.4.1 -- py-1.4.26 -- pytest-2.7.0
rootdir: /Users/docker/Code/pexpect, inifile: setup.cfg
plugins: capturelog, cov
collected 170 items

tests/test_FSM.py .
tests/test_ansi.py .........
tests/test_async.py ....
tests/test_command_list_split.py .
tests/test_constructor.py ..
tests/test_ctrl_chars.py .....
tests/test_destructor.py .
tests/test_dotall.py ..
tests/test_expect.py ..............................
tests/test_filedescriptor.py .....
tests/test_interact.py ...
tests/test_isalive.py .........
tests/test_log.py ....
tests/test_maxcanon.py ...
tests/test_misc.py ............................
tests/test_missing_command.py .
tests/test_performance.py .
tests/test_pickling.py .
tests/test_pxssh.py ...
tests/test_replwrap.py .....
tests/test_repr.py ..
tests/test_run.py .................
tests/test_screen.py ...........
tests/test_timeout_pattern.py .....
tests/test_unicode.py ..........
tests/test_which.py ......
tests/test_winsize.py .
```

3.4.3
-------
```
(pexpect34)C:\Users\docker\Code\pexpect (asyncio-eof)>py.test tests/test_async.py
============================================ test session starts =============================================
platform darwin -- Python 3.4.3 -- py-1.4.26 -- pytest-2.7.0
rootdir: /Users/docker/Code/pexpect, inifile: setup.cfg
plugins: capturelog, cov
collected 4 items

tests/test_async.py .F..

================================================== FAILURES ==================================================
________________________________________ AsyncTests.test_expect_exact ________________________________________

self = <tests.test_async.AsyncTests testMethod=test_expect_exact>

    def test_expect_exact(self):
        p = pexpect.spawn('%s list100.py' % sys.executable)
        assert run(p.expect_exact(b'5', async=True)) == 0
        assert run(p.expect_exact(['wpeok', b'11'], async=True)) == 1
>       assert run(p.expect_exact([b'foo', pexpect.EOF], async=True)) == 1

[...]

PR #287, opened 2015-10-13
Asyncio utf8 issue

jquast:
See Issue #286 -- there is a 2nd *new* error on MacOS for the new testcase, but I think it is the same as #213: a TIMEOUT was received instead of expected EOF.

### DIFF
--- a/pexpect/async.py
+++ b/pexpect/async.py
@@ -41,11 +41,11 @@ class PatternWaiter(asyncio.Protocol):
         spawn._log(s, 'read')
 
         if self.fut.done():
-            spawn.buffer += data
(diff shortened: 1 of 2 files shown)


takluyver:
I think this fix is correct, thanks!

I don't think calling the string `ucs` makes it any clearer, though. That's not a convention for naming unicode strings I've ever seen before, and that C code is using it for a single unicode code point, not a string. Also, if spawn is in bytes mode, `spawn._decoder.decode(data)` will return bytes, in which case calling it `ucs` is completely misleading.

takluyver:
I rewrote the last commit to change the variable name back to s. Merging.

PR #288, opened 2015-10-13
Stop asyncio listening to pty once we've found what we need

takluyver:
Closes gh-213 (fingers crossed)

One symphony and several crazy theories later...

It looks like the problem was the asyncio event loop keeping a reference to the transport, which in turn keeps a reference to the protocol. So after an `expect(... async=True)` call had been made, the transport/protocol set up by that could keep reading events after the expect call had returned. A subsequent `expect(... async=True)` call sets up a new transport/protocol, but it misses events that have already been sent to the previous transport/protocol.

The solution, I hope, is to explicitly detach the transport from the event loop once we've found what we need by calling `transport.pause_reading()`. This makes the tests pass locally on Python 3.4.3.

### DIFF
--- a/pexpect/async.py
+++ b/pexpect/async.py
@@ -23,6 +23,7 @@ def expect_async(expecter, timeout=None):
         return expecter.timeout(e)
 
 class PatternWaiter(asyncio.Protocol):
+    transport = None
     def __init__(self, expecter):
         self.expecter = expecter
         self.fut = asyncio.Future()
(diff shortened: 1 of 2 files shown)


takluyver:
Tests pass on Python 3.5 on Travis :-)

pexpect/async.py
```diff
@@ -30,10 +31,15 @@ def __init__(self, expecter):
     def found(self, result):
         if not self.fut.done():
             self.fut.set_result(result)
+            self.transport.pause_reading()
```
jquast: I should be embarrassed... I am the one who first proposed/contributed the ``transport.pause_reading`` method to the asyncio library :)

jquast:
This is great @takluyver !! I ran into this issue last night and was preparing you new test cases this morning to prove this "stops discovering new data after first pattern matched" issue, I had written code:

```python
    yield from child.expect_exact("telnet> ", async=True)
    child.sendline()
    yield from child.expect_exact("telnet> ", async=True)
(...)
```

where the second expect was always raising timeout -- with this patch, passes! Thank you !

pexpect/async.py
takluyver: Ha! Well, I guess you were pre-emptively ensuring that the machinery to solve our problem was in place ;-)

Issue #289, opened 2015-10-13
pexpect and sagemath


pcpa:
I did package sagemath on fedora some years ago. Right now, the only bundled interface is pexpect.
See https://fedorahosted.org/fpc/ticket/238
The main problem was that, at the time, newer pexpect would not play well with sagemath, there were some echo issues, and getting out of sync.
Just a minor heads up, and letting you know that I would want to help in making sure it works with upstream releases of pexpect.
I remember I did several attempt patches, but never got it fully functional, because I am not 100% familiar with either pexpect or sagemath interface to pexpect.

billpage:
You might be interested in

http://trac.sagemath.org/ticket/10295

and the email thread at:

https://groups.google.com/forum/#!searchin/sage-devel/unicode$20in$20sage$20worksheets/sage-devel/oyTKIfPlCRs/so6s0yO59c0J

as well as the patch at

https://github.com/pexpect/pexpect/pull/216

jdemeyer:
I think this can be closed now: SageMath is using pexpect-4.0.1 + some pull requests (of those, #304 and #291 have not been merged upstream yet).

takluyver:
OK, thanks.

PR #291, opened 2015-10-14
Add new ``delayafterread`` class attribute, for removing Superfluous sleep

jquast:
Problem
-----------

When mixing pexpect with other I/O intensive applications, the releasing of the GIL periodically though technically incorrect, is helpful to prevent a class of downstream developer bugs.  Such "bugs" are subtle race conditions. 

As an example, test_winsize.py fails without such sleep call because its assisting test subprocess, sigwinch_report.py fails, ``RuntimeError: reentrant call inside <_io.BufferedWriter name='<stdout>'>\r\n"`` -- if we look at the code in question, we are in the wrong by performing I/O (print) from within a signal handler.

It is not however obvious at all that this would have anything to do with ``delayaftersleep``, so it is preferable for the project to default to this "safe", but slow behavior.

Relates to #217 #216 #215 #289 

Solution
-----------

We retain existing behavior and provide ``None`` conditional to revert to pexpect 2.0 behavior.

Testing was not done, please request testing by @billpage or @pcpa or any sagemath developer to confirm that this resolves their needs.

References
---------------

http://jessenoller.com/blog/2009/02/01/python-threads-and-the-global-interpreter-lock
https://docs.python.org/2/c-api/init.html#releasing-the-gil-from-extension-code

### DIFF
--- a/pexpect/expect.py
+++ b/pexpect/expect.py
@@ -95,7 +95,6 @@ class Expecter(object):
                     return self.timeout()
                 # Still have time left, so read more data
                 incoming = spawn.read_nonblocking(spawn.maxread, timeout)
-                time.sleep(0.0001)
                 if timeout is not None:
                     timeout = end_time - time.time()
         except EOF as e:
--- a/pexpect/expect.py
+++ b/pexpect/expect.py
@@ -95,6 +95,8 @@ class Expecter(object):
                     return self.timeout()
                 # Still have time left, so read more data
                 incoming = spawn.read_nonblocking(spawn.maxread, timeout)
+                if self.spawn.delayafterread is not None:
+                    time.sleep(self.spawn.delayafterread)
                 if timeout is not None:
                     timeout = end_time - time.time()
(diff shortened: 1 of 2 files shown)


jquast:
needs:
- trivial test of using ``None`` value.
- ok from @takluyver

takluyver:
I don't really like that it's necessary to expose to users hackish details like whether the GIL is released at a specific point in the code - that feels like something that shouldn't matter. But I accept that in this kind of systems programming, it may be necessary to do things like that. So I'm OK with this change if the sage people need it.

Once there's a simple smoke test for setting it to None, we can merge this.

jdemeyer:
> I don't really like that it's necessary to expose to users hackish details like whether the GIL is released at a specific point in the code

For SageMath, the only thing that matters is the performance loss due to the many calls to `time.sleep()`. Whether the GIL is released or not is irrelevant.

takluyver:
Merged via #313

PR #303, opened 2015-12-04
Allow overriding the call to PtyProcess.spawn()

jdemeyer:
For use in SageMath, it is useful to be able to override some methods of the `PtyProcess` class. For this, we need to be able to customize the call `PtyProcess.spawn()` in pexpect.

### DIFF
--- a/pexpect/pty_spawn.py
+++ b/pexpect/pty_spawn.py
@@ -290,7 +290,7 @@ class spawn(SpawnBase):
             self.args = [a if isinstance(a, bytes) else a.encode(self.encoding)
                          for a in self.args]
 
-        self.ptyproc = self.spawnpty(self.args, env=self.env,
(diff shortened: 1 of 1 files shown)


pexpect/pty_spawn.py
```diff
@@ -300,6 +300,10 @@ def preexec_wrapper():
         self.terminated = False
         self.closed = False
 
+    def spawnpty(self, args, **kwargs):
```
takluyver: I'm inclined to make this a private method for now - users should ~never call it, and I think it's only in fairly unusual circumstances that people should be overriding it.

@jquast, thoughts?

jquast: Second -- please rename to ``_spawnpty`` or similar.

jquast:
+1

PR #304, opened 2015-12-04
Read more bytes

jdemeyer:
Optimize `read_nonblocking()` in two ways:
1. When a lot of data is available to read, return all of it. This speeds up callers of `read_nonblocking()` such as `expect_loop()` because there are a lot less iterations in the loop.
2. Only call `is_alive()` if there is no data ready to read. This avoids a lot of pointless calls to `is_alive()`.

Downstream: http://trac.sagemath.org/ticket/10295

### DIFF
--- a/pexpect/pty_spawn.py
+++ b/pexpect/pty_spawn.py
@@ -430,61 +430,83 @@ class spawn(SpawnBase):
         available right away then one character will be returned immediately.
         It will not wait for 30 seconds for another 99 characters to come in.
 
-        This is a wrapper around os.read(). It uses select.select() to
-        implement the timeout. '''
+        On the other hand, if there are bytes available to read immediately,
+        all those bytes will be read (up to the buffer size). So, if the
+        buffer size is 1 megabyte and there is 1 megabyte of data available
+        to read, the buffer will be filled, regardless of timeout.
+
+        This is a wrapper around os.read(). It uses select.select() or
+        select.poll() to implement the timeout. '''
 
         if self.closed:
             raise ValueError('I/O operation on closed file.')
 
+        if self.use_poll:
+            def select(timeout):
+                return poll_ignore_interrupts([self.child_fd], timeout)
+        else:
+            def select(timeout):
+                return select_ignore_interrupts([self.child_fd], [], [], timeout)[0]
+
+        # If there is data available to read right now, read as much as
+        # we can. We do this to increase performance if there are a lot
+        # of bytes to be read. This also avoids calling isalive() too
+        # often. See also:
+        # * https://github.com/pexpect/pexpect/pull/304
+        # * http://trac.sagemath.org/ticket/10295
+        if select(0):
+            try:
+                incoming = super(spawn, self).read_nonblocking(size)
+            except EOF:
+                # Maybe the child is dead: update some attributes in that case
+                self.isalive()
+                raise
+            while len(incoming) < size and select(0):
+                try:
+                    incoming += super(spawn, self).read_nonblocking(size - len(incoming))
+                except EOF:
+                    # Maybe the child is dead: update some attributes in that case
+                    self.isalive()
+                    # Don't raise EOF, just return what we read so far.
+                    return incoming
+            return incoming
+
         if timeout == -1:
             timeout = self.timeout
 
-        # Note that some systems such as Solaris do not give an EOF when
-        # the child dies. In fact, you can still try to read
-        # from the child_fd -- it will block forever or until TIMEOUT.
-        # For this case, I test isalive() before doing any reading.
-        # If isalive() is false, then I pretend that this is the same as EOF.
         if not self.isalive():
-            # timeout of 0 means "poll"
-            if self.use_poll:
-                r = poll_ignore_interrupts([self.child_fd], timeout)
-            else:
-                r, w, e = select_ignore_interrupts([self.child_fd], [], [], 0)
-            if not r:
-                self.flag_eof = True
-                raise EOF('End Of File (EOF). Braindead platform.')
+            # The process is dead, but there may or may not be data
+            # available to read. Note that some systems such as Solaris
+            # do not give an EOF when the child dies. In fact, you can
+            # still try to read from the child_fd -- it will block
+            # forever or until TIMEOUT. For that reason, it's important
+            # to do this check before calling select() with timeout.
+            if select(0):
+                return super(spawn, self).read_nonblocking(size)
+            self.flag_eof = True
+            raise EOF('End Of File (EOF). Braindead platform.')
         elif self.__irix_hack:
             # Irix takes a long time before it realizes a child was terminated.
+            # Make sure that the timeout is at least 2 seconds.
             # FIXME So does this mean Irix systems are forced to always have
             # FIXME a 2 second delay when calling read_nonblocking? That sucks.
-            if self.use_poll:
-                r = poll_ignore_interrupts([self.child_fd], timeout)
-            else:
-                r, w, e = select_ignore_interrupts([self.child_fd], [], [], 2)
-            if not r and not self.isalive():
-                self.flag_eof = True
-                raise EOF('End Of File (EOF). Slow platform.')
-        if self.use_poll:
-            r = poll_ignore_interrupts([self.child_fd], timeout)
-        else:
-            r, w, e = select_ignore_interrupts(
-                [self.child_fd], [], [], timeout
-            )
+            if timeout is not None and timeout < 2:
+                timeout = 2
 
-        if not r:
-            if not self.isalive():
-                # Some platforms, such as Irix, will claim that their
-                # processes are alive; timeout on the select; and
-                # then finally admit that they are not alive.
-                self.flag_eof = True
-                raise EOF('End of File (EOF). Very slow platform.')
-            else:
-                raise TIMEOUT('Timeout exceeded.')
-
-        if self.child_fd in r:
+        # Because of the select(0) check above, we know that no data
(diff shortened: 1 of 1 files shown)


jdemeyer:
After thinking about this a bit more, I'm wondering if I should instead patch `read_nonblocking()` itself instead of the `except_loop()` front-end. That would probably be even better, since one could avoid many calls to `is_alive` yielding yet another (although relatively small) speed-up.

jdemeyer:
If you need to be convinced, some timings from SageMath, see http://trac.sagemath.org/ticket/10295

pexpect 2.0 upstream:
```
sage: %time _ = str(gp("2^2^22"))
CPU times: user 184 ms, sys: 4 ms, total: 188 ms
Wall time: 384 ms
```

pexpect 4.0.1 upstream:
```
sage: %time _ = str(gp("2^2^22"))
CPU times: user 308 ms, sys: 9 ms, total: 317 ms
Wall time: 571 ms
```

pexpect 4.0.1 + #291 + #304:
```
sage: %time _ = str(gp("2^2^22"))
CPU times: user 16 ms, sys: 6 ms, total: 22 ms
Wall time: 216 ms
```

Note in particular the CPU time, which is reduced by more than a factor 10.

jdemeyer:
Like I said, this new commit makes the changes to `read_nonblocking()` itself. I think this is a simpler and better solution.

pexpect/pty_spawn.py
```diff
@@ -430,11 +452,7 @@ def read_nonblocking(self, size=1, timeout=-1):
         # For this case, I test isalive() before doing any reading.
         # If isalive() is false, then I pretend that this is the same as EOF.
```
takluyver: The comments suggest that without this check, there's a risk of blocking on Solaris. So I'd rather not bypass it unless we're confident that this is no longer an issue when you run in Solaris.

billpage: Are you suggesting that Solaris might block even when calling __select with a timeout of 0?  It looks to me like these comments must refer to an earlier version of this code where timeout may not have been 0.

takluyver: In general, on EOF, select will indicate that there is something to be read, and then - depending on the system - read will either fail or return an empty string. This comment appears to be saying that on Solaris, read on a pty after EOF will block, so we can only detect EOF by checking if the child process has gone. If so, by reading before this check, we could get blocked.

It may be that that is outdated, and Solaris has since been fixed to do something sane. But I don't want to throw away hard won compatibility code until we're confident that it's no longer needed. @jquast has been setting up a greater diversity of different Unix systems than I have ready access to - he may be able to check it.

jdemeyer: > This comment appears to be saying that on Solaris, read on a pty after EOF will block

The question is really: what will `select()` do in this situation? If `select()` simply times out, we are safe.

jdemeyer: > This comment appears to be saying that on Solaris, read on a pty after EOF will block

Actually, if you read this code:
```python
       if not self.isalive():
            # timeout of 0 means "poll"
            r, w, e = self.__select([self.child_fd], [], [], 0)
            if not r:
                self.flag_eof = True
                raise EOF('End Of File (EOF). Braindead platform.')
```
you see that `select()` indicates that `child_fd` is not ready to read. So I think that the new code is safe.

takluyver: Yes, you're right, sorry.

takluyver:
I'm having a bit of trouble understanding why this would make a significant difference. My understanding is that if there are `size` bytes ready to read, when the parent class calls `os.read(fd, size)`, it should return `size` bytes at once. So why do we need to loop and read more data until we reach size?

I thought I had understood it earlier, but now I can't work it out. It may just be that I'm tired.

jdemeyer:
> if there are size bytes ready to read, when the parent class calls os.read(fd, size), it should return size bytes at once

At least on Linux, this is not the case. It might be true for other kind of files but not for pty's.

jquast:
looking ok so far. I'll do some work shortly to run tests on non-linux systems.

As for,

> when the parent class calls os.read(fd, size), it should return size bytes at once

``read(2)`` has the right to return less bytes than requested, this is especially true on sockets, pty's, and other special files.

From http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man2/pread.2?query=read

> Upon successful completion, read() ...  guarantees to read the number of bytes requested **if the descriptor references a normal file** that has that many bytes left before the end-of-file, **but in no other case**.

This is somewhat parroted by weasel wording in python's documentation:

> Read **at most** n bytes

takluyver:
Fair enough, but I'd still assume that if you ask for n bytes and there are >= n ready in the buffer, it will most likely return n, even if it's not guaranteed. Any idea why it wouldn't do that?

jquast:
> it will most likely return n, even if it's not guaranteed. Any idea why it wouldn't do that?

Given: A child process outputting 1 byte to stdout, then ``sleep(1)`` in loop.
If ``os.read(child_fd, 50)`` is called, it may return only 1 byte. The end of stream is reached, but not end of file.

tests/test_expect.py
```diff
@@ -433,7 +433,7 @@ def test_before_after_exact(self):
         self._before_after(p)
 
     def _ordering(self, p):
-        p.timeout = 5
+        p.timeout = 20
```
jquast: I presume this small test case change is not necessary, may it be reverted?

pexpect/pty_spawn.py
```diff
@@ -415,14 +415,36 @@ def read_nonblocking(self, size=1, timeout=-1):
...
+            while len(incoming) < size:
+                r, w, e = self.__select([self.child_fd], [], [], 0)
+                if not r:
+                    break
+                try:
+                    incoming += super(spawn, self).read_nonblocking(size - len(incoming))
+                except EOF:
+                    # Ignore EOF, just return the bytes we got so far.
+                    break
+            return incoming
```
jquast: This is good. I support this "avoid busy loop" style of optimization.

pexpect/pty_spawn.py
```diff
@@ -430,11 +452,7 @@ def read_nonblocking(self, size=1, timeout=-1):
         # For this case, I test isalive() before doing any reading.
         # If isalive() is false, then I pretend that this is the same as EOF.
         if not self.isalive():
-            # timeout of 0 means "poll"
-            r, w, e = self.__select([self.child_fd], [], [], 0)
-            if not r:
-                self.flag_eof = True
-                raise EOF('End Of File (EOF). Braindead platform.')
```
jquast: The removal of code here is correct as such unconditional 0-second check is done.

However ``isalive`` is now deferred until after ``select()``. This might be a problem.

pexpect/pty_spawn.py
```diff
@@ -456,10 +480,7 @@ def read_nonblocking(self, size=1, timeout=-1):
             else:
                 raise TIMEOUT('Timeout exceeded.')
 
-        if self.child_fd in r:
-            return super(spawn, self).read_nonblocking(size)
-
-        raise ExceptionPexpect('Reached an unexpected state.')  # pragma: no cover
```
jquast: good

jquast:
Two strange EOF related errors on Linux and MacOS.

Linux, https://teamcity-master.pexpect.org/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=16722#_focus=1960

MacOS, https://teamcity-master.pexpect.org/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=16723#_focus=1846

Branching and experimenting with reorder if ``isalive()``.

pexpect/pty_spawn.py
```diff
@@ -430,11 +452,7 @@ def read_nonblocking(self, size=1, timeout=-1):
         # For this case, I test isalive() before doing any reading.
         # If isalive() is false, then I pretend that this is the same as EOF.
         if not self.isalive():
-            # timeout of 0 means "poll"
-            r, w, e = self.__select([self.child_fd], [], [], 0)
-            if not r:
-                self.flag_eof = True
```
jquast: ``flag_eof`` was missed, added 5e6e2a1d01ef1943b69217d07f4202268b1b2a78

tests/test_expect.py
jdemeyer: Well, I increased this because there was a timeout on one of the bots.

I think 5 seconds is too short, you will get timeouts now and then if the machine happens to be busy when running the test.

jquast:
I'm ok with the #309 variant of this PR that I have proposed. I ran it through MacOS, Linux, and FreeBSD testing several times successfully. @takluyver?

tests/test_expect.py
jdemeyer: I got exactly the same false-positive failure again on #313. Please keep this change instead of reverting it!

jdemeyer:
Do we still care about the IRIX hack? I'd rather just drop that.

jdemeyer:
Rebased to latest master.

Red-M:
I don't have anything to say for the removal of the IRIX piece here other than what has already been said.

jdemeyer:
> Anyway, would you like to be a co-maintainer of Pexpect? You seem to be more interested in using it and developing it than I am.

Sure. I'm mainly interested because SageMath uses it.

takluyver:
OK, I've invited you to collaborate on the repository. Sorry for the delay.

jdemeyer:
So, do you have any objections to merging this? I know I can merge it myself now, but I wanted to know your opinion.

takluyver:
Sure, go for it.

PR #307, opened 2015-12-09
Allow delaybeforesend=None to skip the sleep completely

jdemeyer:
Yet another micro-optimization.

### DIFF
--- a/doc/commonissues.rst
+++ b/doc/commonissues.rst
@@ -47,7 +47,7 @@ not be an issue for most users. For some applications you might with to turn it
 off::
 
     child = pexpect.spawn ("ssh user@example.com")
-    child.delaybeforesend = 0
+    child.delaybeforesend = None
 
 Truncated output just before child exits
 ----------------------------------------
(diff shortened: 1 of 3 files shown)


takluyver:
1. Is there actually any measurable benefit to avoiding the call of `time.sleep(0)`?
2. Why not just check `== 0` rather than introducing a special alternative value?

jdemeyer:
> Is there actually any measurable benefit to avoiding the call of time.sleep(0)?

I haven't actually tested it. But I know from general experience that avoiding system calls is a good idea: they involve context switches between user-space and kernel-space. This takes a small but measurable amount of time.

> Why not just check == 0 rather than introducing a special alternative value?

I just did that for consistency with #291.

jquast:
The purpose of a sleep call for 0 seconds would be to allow other threads to run,
https://github.com/python/cpython/blob/master/Modules/timemodule.c#L1431

jquast:
I support this for merge, ok @takluyver?

The ``None`` type check, rather than a 0-second check is significant, and a value of 0 continues to release the thread (matches previous behavior that may be dependent)

jdemeyer:
Tests added.

What does `needs-changelog` mean, what should I change?

jquast:
> What does needs-changelog mean, what should I change?

We will probably do that for you, It is a reminder to to update doc/history.rst for a 4.1 release, to describe this new keyword argument behavior.

takluyver:
Thanks, this looks fine to me. I don't especially like passing the gil-releasing decision on to users - it feels like a detail they shouldn't have to care about - but I guess that's the way things go with this type of code.

PR #308, opened 2015-12-10
Call send() just once in sendline()

jdemeyer:
Optimize `sendline()` by calling `send()` just once.

### DIFF
--- a/pexpect/pty_spawn.py
+++ b/pexpect/pty_spawn.py
@@ -525,10 +525,7 @@ class spawn(SpawnBase):
         written.  Only a limited number of bytes may be sent for each
         line in the default terminal mode, see docstring of :meth:`send`.
         '''
-
-        n = self.send(s)
-        n = n + self.send(self.linesep)
-        return n
+        return self.send(s + self.linesep)
 
     def _log_control(self, s):
         """Write control characters to the appropriate log files"""


jdemeyer:
There seems to be a `unicode` vs. `bytes` issue on Python 3. I must say that I don't know much about this stuff (I never use Python 3).

Why do you have `self._coerce_send_string(s)` and `self._encoder.encode(s, final=False)` in the `send()` method? Is it important that the `log()` happens in between those two?

takluyver:
The data being sent should be logged in the native type of the spawn instance (`bytes` if encoding is None, otherwise unicode (str on Python 3)). For backwards compatibility, a spawn instance in bytes mode also accepts unicode, so the call to _coerce_send_string normalises it.

 Then, to actually send the data, it always needs to be converted to bytes, which is what the call to `_encoder.encode` does.

jquast:
Closing as discussed,

``send()`` **must** be called for ``self.linesep`` separately.  ``a + b`` in this call would have an implicit up conversion of bytes to unicode using the ASCII encoding. Python 3 correctly guards this condition all together with ``TypeError: Can't convert 'bytes' object to str implicitly``

jdemeyer:
A fixed version is up at #311.

PR #306, opened 2015-12-08
Add bashrc.sh to package_data

takluyver:
Closes #305

PR #309, opened 2015-12-13
Reduce time of read_nonblocking() for bulk output

jquast:
Follows up pull #304

### DIFF
--- a/tests/test_interact.py
+++ b/tests/test_interact.py
@@ -24,6 +24,7 @@ from __future__ import unicode_literals
 
 import os
 import pexpect
+import platform
 import unittest
 import sys
 from . import PexpectTestCase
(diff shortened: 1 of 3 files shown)


tests/test_interact.py
```diff
@@ -62,10 +63,12 @@ def test_interact_escape_None(self):
         p.sendcontrol(']')
         p.expect('29<STOP>')
         p.send('\x00')
-        if not os.environ.get('TRAVIS', None):
-            # on Travis-CI, we sometimes miss trailing stdout from the
-            # chain of child processes, not entirely sure why. So this
-            # is skipped on such systems.
+
+        if not platform.system() == 'Linux':
```
takluyver: I'm not quite happy that this test piece which was previously skipped only on Travis is now skipped on all Linux systems. We presumably had it passing on Linux at some point in the past. Is that no longer the case even without the changes in this PR?

takluyver: @jquast I'd like to get this test question sorted out, merge this and cut a new release (I guess there's enough changes to call it 4.1). I'd like to get the fix from #306 out soon, so if we need longer to work this out, I may cut a new release without it.

PR #310, opened 2015-12-13
Bugfix: AttributeError: 'spawn' object has no attribute 'ptyproc'

### LINKED ISSUE #298: AttributeError: 'my_spawn' object has no attribute 'ptyproc'
We subclass `spawn` from `pty_spawn.py` to defer expecting on commands through a telnet connection. In this case, the command is not know at init-time, and thus `command` is `None`, which in turn avoids running `_spawn()` in the super class.

In this setup, a timeout exception tries to stringify the `spawn` class, which reads the `flag_eof` property, where `self.ptyproc` is accessed, but doesn't exist (due to not calling `_spawn()`, leading to an uncaught exception:
```
  File "./except.py", line 19, in <module>
    child.expect('foo', 1)
  File "/usr/lib/python2.7/site-packages/pexpect/spawnbase.py", line 315, in expect
    timeout, searchwindowsize, async)
  File "/usr/lib/python2.7/site-packages/pexpect/spawnbase.py", line 339, in expect_list
    return exp.expect_loop(timeout)
  File "/usr/lib/python2.7/site-packages/pexpect/expect.py", line 95, in expect_loop
    return self.timeout()
  File "/usr/lib/python2.7/site-packages/pexpect/expect.py", line 65, in timeout
    msg = str(spawn)
  File "/usr/lib/python2.7/site-packages/pexpect/pty_spawn.py", line 213, in __str__
    s.append('flag_eof: ' + str(self.flag_eof))
  File "/usr/lib/python2.7/site-packages/pexpect/pty_spawn.py", line 569, in flag_eof
    return self.ptyproc.flag_eof
AttributeError: 'my_spawn' object has no attribute 'ptyproc'
```

The exception is reproducible with the following code:
```
 #!/usr/bin/env python2

import pexpect

class my_spawn(pexpect.spawn):
    def __init__(self, command=None):
        self.args = None
        self.command = command
        pexpect.spawn.__init__(self, None, None, timeout=1)

        self.child_fd = -1
        self.own_fd = False
        self.closed = False

    def read_nonblocking (self, size=1, timeout=-1):
        return ""

child = my_spawn('sleep 10')
child.expect('foo', 1)
```

jquast:
When subclassing ``spawn()``, or otherwise using ``command=None`` argument as in the given test, ``self.ptyproc`` is left unassigned resulting in exception, ``AttributeError: 'spawn' object has no attribute 'jptyproc'`` instead of ``TIMEOUT``.

Solution is to wrap access of this variable by ``hasattr``. Closes #298

### DIFF
--- a/pexpect/pty_spawn.py
+++ b/pexpect/pty_spawn.py
@@ -210,7 +210,8 @@ class spawn(SpawnBase):
         s.append('match: %r' % (self.match,))
         s.append('match_index: ' + str(self.match_index))
         s.append('exitstatus: ' + str(self.exitstatus))
-        s.append('flag_eof: ' + str(self.flag_eof))
+        if hasattr(self, 'ptyproc'):
+            s.append('flag_eof: ' + str(self.flag_eof))
         s.append('pid: ' + str(self.pid))
         s.append('child_fd: ' + str(self.child_fd))
         s.append('closed: ' + str(self.closed))
--- a/tests/test_repr.py
+++ b/tests/test_repr.py
@@ -28,7 +28,7 @@ class TestCaseMisc(PexpectTestCase.PexpectTestCase):
         """ Exercise derived spawn.__str__() """
         # given,
         child = pexpect.spawn(None, None)
-        child.closed = False
+        child.read_nonblocking = lambda size, timeout: b''
         try:
             child.expect('alpha', timeout=0.1)
         except pexpect.TIMEOUT:
--- a/tests/test_repr.py
+++ b/tests/test_repr.py
@@ -31,7 +31,7 @@ class TestCaseMisc(PexpectTestCase.PexpectTestCase):
         child.read_nonblocking = lambda size, timeout: b''
         try:
             child.expect('alpha', timeout=0.1)
-        except pexpect.TIMEOUT:
-            pass
+        except pexpect.TIMEOUT as e:
+            str(e)  # Smoketest
         else:
-            assert False, 'TIMEOUT exception expected. No exception aised.'
+            assert False, 'TIMEOUT exception expected. No exception raised.'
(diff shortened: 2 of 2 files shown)


pexpect/pty_spawn.py
```diff
@@ -210,7 +210,8 @@ def __str__(self):
         s.append('match: %r' % (self.match,))
         s.append('match_index: ' + str(self.match_index))
         s.append('exitstatus: ' + str(self.exitstatus))
-        s.append('flag_eof: ' + str(self.flag_eof))
+        if hasattr(self, 'ptyproc'):
+            s.append('flag_eof: ' + str(self.flag_eof))
```
takluyver: This feels distinctly band-aid-ish. I'm happy for it to go in so that other errors show up as expected, but we should probably look more closely at whether `self.flag_eof` can do something sensible in such subclasses.

Even thinking about thinking about that makes me sleepy ;-)

tests/test_repr.py
```diff
@@ -24,3 +24,14 @@ def test_str_spawn(self):
...
+        """ Exercise derived spawn.__str__() """
+        # given,
+        child = pexpect.spawn(None, None)
+        child.read_nonblocking = lambda size, timeout: b''
+        try:
+            child.expect('alpha', timeout=0.1)
+        except pexpect.TIMEOUT:
+            pass
+        else:
+            assert False, 'TIMEOUT exception expected. No exception aised.'
```
takluyver: aised -> raised. Will fix and merge.

tests/test_repr.py
```diff
@@ -24,3 +24,14 @@ def test_str_spawn(self):
...
         assert isinstance(value, str)
 
+    def test_str_before_spawn(self):
+        """ Exercise derived spawn.__str__() """
+        # given,
+        child = pexpect.spawn(None, None)
+        child.read_nonblocking = lambda size, timeout: b''
+        try:
+            child.expect('alpha', timeout=0.1)
+        except pexpect.TIMEOUT:
```
takluyver: Actually, I don't think this test case would exhibit the failure that you mentioned, because it arises when something tries to *display* the error, and if you catch it, nothing will do that.

PR #313, opened 2016-01-05
Add new ``delayafterread`` class attribute, for removing Superfluous sleep

jdemeyer:
The many calls to `time.sleep(0.0001)` in pexpect cause a serious performance loss when using pexpect in SageMath: that sleep was the main reason that SageMath was using pexpect-2.0 for a long time, see #215. And it seems that the releasing of the GIL during this sleep can also problematic, see #291.

To solve this problem, we make the sleep optional with a new `delayafterread` class attribute: if set to `None`, the sleep will be skipped completely (which is not the same as calling `time.sleep(0)`).

### DIFF
--- a/pexpect/expect.py
+++ b/pexpect/expect.py
@@ -95,7 +95,8 @@ class Expecter(object):
                     return self.timeout()
                 # Still have time left, so read more data
                 incoming = spawn.read_nonblocking(spawn.maxread, timeout)
-                time.sleep(0.0001)
+                if self.spawn.delayafterread is not None:
+                    time.sleep(self.spawn.delayafterread)
                 if timeout is not None:
                     timeout = end_time - time.time()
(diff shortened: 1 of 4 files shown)


takluyver:
I'm confused - what's the relationship between this and #291, which appears to have precisely the same title?

jdemeyer:
> I'm confused - what's the relationship between this and #291, which appears to have precisely the same title?

This pull request supersedes #291.

As far as I know, it is not possible in GitHub for me to push to the branch from #291. So if I want to add commits, I need to create a completely new pull request.

jdemeyer:
To elaborate a bit more: at #291, you asked for a test to be added. This pull request adds such a test (together with a few trivial changes).

takluyver:
Ah, gotcha. Thanks!

PR #320, opened 2016-01-22
Provide example use of exitstatus and signalstatus

nchammas:
Looking through [the doc](http://pexpect.readthedocs.org/en/stable/api/pexpect.html#spawn-class) several times, I missed the details of how to get the exit code of a spawned process.

Normally, important class attributes like `exitstatus` and `signalstatus` are called out in the docs with some kind of special formatting. Since I don't know how to do that here, I opted instead to add an example.

### DIFF
--- a/pexpect/pty_spawn.py
+++ b/pexpect/pty_spawn.py
@@ -154,7 +154,12 @@ class spawn(SpawnBase):
         in self.exitstatus or self.signalstatus. If the child exited normally
         then exitstatus will store the exit return code and signalstatus will
         be None. If the child was terminated abnormally with a signal then
-        signalstatus will store the signal value and exitstatus will be None.
+        signalstatus will store the signal value and exitstatus will be None::
+
+            child = pexpect.spawn('some_command')
+            child.close()
+            print(child.exitstatus, child.signalstatus)
+
         If you need more detail you can also read the self.status member which
         stores the status returned by os.waitpid. You can interpret this using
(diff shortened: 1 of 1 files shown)


nchammas:
To be clear, the root issue I'm looking to address here is that `exitstatus` and `signalstatus` aren't sufficiently called out in the docs. Providing an example of how to use them is one possible way to address that.

jquast:
Thank you @nchammas we'll review a way to make this api information more easily found in docs.

nchammas:
:+1: And I'm fine with closing this PR if y'all find a way to do this that's more "proper".

pexpect/pty_spawn.py
```diff
@@ -155,6 +155,11 @@ def __init__(self, command, args=[], timeout=30, maxread=2000,
         then exitstatus will store the exit return code and signalstatus will
         be None. If the child was terminated abnormally with a signal then
         signalstatus will store the signal value and exitstatus will be None.
+        
+            child = pexpect.spawn('some_command')
```
takluyver: I think this looks helpful, but can you stick a `::` above it, with a blank line either side? That will make it show up as a code block [in the docs](http://pexpect.readthedocs.org/en/stable/api/pexpect.html#spawn-class).

nchammas: Done.

takluyver:
Thanks. Merging this as an improvement; we can always improve more later.

PR #323, opened 2016-01-29
Use setuptools, not distutils

njwhite:
To make wheel-building easier (see [this](http://stackoverflow.com/a/27868004/42543)).


takluyver:
I prefer to avoid setuptools. It has messed my computer up so many times when I've installed something that uses it.

PR #327, opened 2016-02-14
Use flit instead of setuptools for ptyprocess  (Teamcity CI).

jquast:
This allows us to continue to test ptyprocess feature branches using pexpect,
introducing the new function `make_wheel_for_ptyprocess` and amending
existing code into `prepare_environment`, which installs the wheel file created
by flit.

this has gotten more complex, so we have made individual function units to
hopefully make it a bit more straight-forward to stumble into.

Note this does not effect our package, requires no version bump, release, or changelog item. This is used only by the mac/linux/freebsd build slaves

### DIFF
--- a/tools/teamcity-runtests.sh
+++ b/tools/teamcity-runtests.sh
@@ -14,7 +14,7 @@ export PYTHONIOENCODING=UTF8
 export LANG=en_US.UTF-8
 
 pyversion=$1
-version_flit=3.5
+pyversion_flit=3.5
 shift
 here=$(cd `dirname $0`; pwd)
 osrel=$(uname -s)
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks for working out how to do this! I can't deny that this is a fair bit of extra complexity for our custom test script, though. I'm obviously rather fond of my own packaging tool, but I won't be offended if you'd rather merge pexpect/ptyprocess#32 and go back to the conventional tools until/unless a new standard is hammered out. The intent is not to make things like this more complex!

takluyver:
PS I'm turning in for the night on this side of the pond, so I won't see what happens until tomorrow morning. Enjoy your long weekend!

takluyver:
It may affect this decision that there are a couple of PEPs in the works (516, 517) to define new sdist formats, which flit would support. Neither of them are moving very quickly, though.

PR #335, opened 2016-03-03
change passmass example to not log writes

joelcollinsdc:


takluyver:
Sorry this has sat around, merging now.

PR #340, opened 2016-04-05
Added support for sockets.

MountainRider:
From the maintainer:

    We changed the inheritance around - in earlier versions of
    pexpect, fdspawn inherited from spawn, even though spawn is
    actually the more specific case (it needs to know about the
    bizarreness of ptys). Now spawn and fdspawn both inherit from
    spawnbase. This makes a lot more sense, but it may well be
    causing what you see. I'd suggest you make a separate subclass
    for dealing with sockets.

Added a socket_pexpect module which creates a socket_spawn class.
The socket_spawn class overrides the read_nonblocking method of
the fdspawn class. The read_nonblocking method uses select to
implement the timeout and check if there is anything to read from
the socket. If there is something to read, the superclass
read_nonblocking is called.

### DIFF
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -9,6 +9,7 @@ API documentation
    popen_spawn
    replwrap
    pxssh
+   socket_pexpect
 
 The modules ``pexpect.screen`` and ``pexpect.ANSI`` have been deprecated in
 Pexpect version 4. They were separate from the main use cases for Pexpect, and
--- a/doc/api/socket_pexpect.rst
+++ b/doc/api/socket_pexpect.rst
@@ -1,0 +1,16 @@
+socket_pexpect - use pexpect with a socket
+==========================================
+
+.. automodule:: pexpect.socket_pexpect
+
+socket_pexpect class
+--------------------
+
+.. autoclass:: socket_spawn
+   :show-inheritance:
+
+   .. method:: expect
+               expect_exact
+               expect_list
+
+      As :class:`pexpect.spawn`.
--- a/pexpect/socket_pexpect.py
+++ b/pexpect/socket_pexpect.py
@@ -70,7 +70,7 @@ class socket_spawn(fdspawn):
                     if timeout is not None:
                         timeout = end_time - time.time()
                         if timeout < 0:
-                            return tuple([], [], [])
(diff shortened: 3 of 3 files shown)


MountainRider:
I have fixed a defect in the __select method in the socket_spawn class and I have added some unit tests. The unit tests should fix the code coverage decrease.

takluyver:
This is a totally reasonable thing to do, but I think Pexpect's API is awkward, and it's so bound by backwards compatibility that it's tricky to do anything new with it.

I wonder if we should stop adding new features to Pexpect altogether, maintain it only with bugfixes, and encourage people to build a new package to do the same core task (read from a stream until a certain pattern appears).

MountainRider:
I am happy to help maintain socket support in Pexpect. What is your plan to "encourage people to build a new package"? Do you have an API or abstract class interface in mind?

jquast:
> Do you have an API or abstract class interface in mind?

@MountainRider: We do not have any hidden design document or similar for a "pexpect next generation". Mainly public discussions here in github issues.

> What is your plan to "encourage people to build a new package"?

Personally, I don't have time available in the near future to author a "pexpect next generation" project, but I should hope we could collect some thoughts for any interested parties, perhaps in a design document somewhere.

@nastevens Published a very well-reasoned "pexpect next generation" project as https://github.com/digidotcom/python-streamexpect -- I would encourage anyone to take a look, as this project focuses on expecting "streams", and removes the pseudo-terminal specific cruft.

Unfortunately it does not include asyncio support, I should hope a "next generation stream expect"  library with strong asyncio support should survive the next decades. If I could ever find the time, I would begin with contributing for https://github.com/digidotcom/python-streamexpect/issues/1

takluyver:
Streamexpect has at least a couple of nice things that I'm happy to see, just from looking at the README:

- It takes arbitrary stream objects, including files, sockets and serial ports, without requiring a separate class for each.
- Expect methods return a match object, rather than returning an index and changing the state of the object they're called on.

nastevens:
Thanks @jquast for the callout. I hope that anyone reading this will consider contributing to streamexpect - all skill levels are welcome and I am happy to work with anyone who'd like to take ownership of an open issue.

PR #343, opened 2016-04-06
Add support for sockets

MountainRider:
I am adding support for sockets because we use pexpect heavily and I do not want to have to patch every installation. I tried to follow the naming conventions and coding style in the existing code. Please let me know if I need to make any changes.

### DIFF
--- a/doc/api/index.rst
+++ b/doc/api/index.rst
@@ -9,6 +9,7 @@ API documentation
    popen_spawn
    replwrap
    pxssh
+   socket_pexpect
 
 The modules ``pexpect.screen`` and ``pexpect.ANSI`` have been deprecated in
 Pexpect version 4. They were separate from the main use cases for Pexpect, and
--- a/doc/api/socket_pexpect.rst
+++ b/doc/api/socket_pexpect.rst
@@ -1,0 +1,16 @@
+socket_pexpect - use pexpect with a socket
+==========================================
+
+.. automodule:: pexpect.socket_pexpect
+
+socket_pexpect class
+--------------------
+
+.. autoclass:: socket_spawn
+   :show-inheritance:
+
+   .. method:: expect
+               expect_exact
+               expect_list
+
+      As :class:`pexpect.spawn`.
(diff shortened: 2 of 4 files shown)


MountainRider:
I forgot to remove a temporary import I was using for testing.

MountainRider:
Apparently byte arrays are needed for tests to pass in Python 3.

MountainRider:
Should have known that calling replace on a byte array requires byte array parameters.

MountainRider:
The code coverage decrease is in pty_spawn.py. Since a socket is not a pty, I am not sure I can prevent the decrease because I cannot call isatty.

MountainRider:
I should have read the test more closely. I can call isatty because the check is if it returns False.

MountainRider:
I am not sure how to fix that coverage decrease. Is that caused by some other change?

MountainRider:
After tracing through the code, the only difference between my tests and the fdpexpect tests was that I was not expecting EOF. I ask the other end to disconnect by going into the menu and selecting exit. After that, I can expect EOF.

MountainRider:
Added a test that runs the pexpect session in a sub-process and sends interrupts to it every 250 milliseconds. This should increase the coverage by exercising the code in the __select method.

MountainRider:
Added test for file descriptor not being an integer and for not being a valid file descriptor.

MountainRider:
The pypy test_fileobj failed, but I don't understand why.

MountainRider:
Join the process so that we know it has finished before running any other tests.

MountainRider:
The sub-process in the interrupt test was failing silently.

MountainRider:
I think this is ready for review.

MountainRider:
I copied the __select method from pty_spawn.py into my socket_expect.py. I tried to create a unit test that got 100% coverage of this method, but I could not interrupt the select.select call. I have two questions:
1. Should the __select method be moved into utils.py?
2. Do we even need to handle the possibility that select.select might be interrupted?

Issue #345, opened 2016-04-25
pexpect 4.0.1  - searcher in output is NULL


novakmi:
Looks like searcher value reported in failed `expect` is always NULL from 4.0.x.

Run following script to reproduce:

<pre>
#!/usr/bin/python
import pexpect
child = pexpect.spawn('ls -l')
child.expect('total')
print "searcher as string %s" % str(child.searcher)
print
print "searcher as rep %r" % child.searcher
print
child2 = pexpect.spawn('ls -l')
child2.expect('total2')`
</pre>

With pexpect 3.3 the output is:

<pre>
searcher as string searcher_re:
    0: re.compile("total")

searcher as rep &lt;pexpect.searcher_re object at 0x7f127ec6b2d0&gt;
....
version: 3.3
command: /bin/ls
args: ['/bin/ls', '-l']
searcher: &lt;pexpect.searcher_re object at 0x7f127ec6b310&gt;
buffer (last 100 chars): ''
before (last 100 chars): 'total 4\r\n-rwxrwxr-x 1 user user 251 Apr 25 14:51 pexpt_test.py\r\n'
after: &lt;class 'pexpect.EOF'&gt;
match: None
</pre>

with pexpect 4.0.1 the output is

<pre>
searcher as string None

searcher as rep None
....
command: /bin/ls
args: ['/bin/ls', '-l']
searcher: None
buffer (last 100 chars): ''
before (last 100 chars): 'total 4\r\n-rwxrwxr-x 1 user user 251 Apr 25 14:51 pexpt_test.py\r\n'
after: &lt;class 'pexpect.exceptions.EOF'&gt;
match: None
match_index: None
exitstatus: 0
flag_eof: True
pid: 32087
</pre>

* As you can see, the `searcher` in 4.0.1 is always NULL and is not reported.

In addition:

* For 3.3. the sercher is printed as object `%r`, which does not give user much information, it should be printed as `str()` so user can see what was expected (see `print` commands in the example script)
* For 4.0.1 `pexpect` version is not printed anymore  (see `version: 3.3.` for 3.3. output)

takluyver:
This is because `searcher` is no longer set as an attribute on `spawn`. And it's exactly the kind of reason that I want to start from scratch and create an alternative to Pexpect which doesn't need to adhere to these kinds of historical details.

novakmi:
[...]

PR #346, opened 2016-04-25
Display current searcher in error messages

takluyver:
Without attaching it to the spawn object

Closes gh-345


jquast:
Thanks, feel free to version bump and release to pypi any time.

takluyver:
I have released 4.1.0. I have added you as a package owner on PyPI as well - I had forgotten that that wasn't already the case.

PR #348, opened 2016-05-13
Use PATH from env argument to pexpect.spawn

bz2:
Currently pexpect.spawn takes an env argument along the lines of
subprocess.Popen but fails to use PATH from that env when finding
the binary to run.

The result when running a binary not on the PATH of the parent
process is pexpect will raise an exception, or worse, find a
different binary with the same name.

This change passes the env from spawn into the which function
so the correct PATH is searched and adds test coverage.

### DIFF
--- a/pexpect/pty_spawn.py
+++ b/pexpect/pty_spawn.py
@@ -266,7 +266,7 @@ class spawn(SpawnBase):
             self.args.insert(0, command)
             self.command = command
 
-        command_with_path = which(self.command)
+        command_with_path = which(self.command, env=self.env)
         if command_with_path is None:
             raise ExceptionPexpect('The command was not found or was not ' +
                     'executable: %s.' % self.command)
--- a/pexpect/utils.py
+++ b/pexpect/utils.py
@@ -31,7 +31,7 @@ def is_executable_file(path):
     return os.access(fpath, os.X_OK)
 
 
-def which(filename):
+def which(filename, env=None):
     '''This takes a given filename; tries to find it in the environment path;
     then checks if it is executable. This returns the full path to the filename
     if found and executable. Otherwise this returns None.'''
@@ -39,10 +39,11 @@ def which(filename):
     # Special case where filename contains an explicit path.
     if os.path.dirname(filename) != '' and is_executable_file(filename):
         return filename
-    if 'PATH' not in os.environ or os.environ['PATH'] == '':
+    if env is None:
+        env = os.environ
+    p = env.get('PATH')
+    if not p:
         p = os.defpath
-    else:
-        p = os.environ['PATH']
     pathlist = p.split(os.pathsep)
     for path in pathlist:
         ff = os.path.join(path, filename)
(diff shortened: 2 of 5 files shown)


takluyver:
This technically breaks backwards compatibility, because something may be relying on the command being looked up in the parent process' PATH, not the subprocess'. Maybe that should be considered a corner case, though Pexpect is old enough that I'm wary about changing stuff.

Have you checked what subprocess does?

@jquast , what's your preference?

bz2:
The subprocess module gets it right:

```
>>> import subprocess
>>> subprocess.call("a-script.sh", env={"PATH": "/tmp"})
Running script not on $PATH
0
```

Current pexpect:

```
>>> import pexpect
>>> pexpect.run("a-script.sh", env={"PATH": "/tmp"})
  ...
pexpect.exceptions.ExceptionPexpect: The command was not found or was not executable: a-script.sh.
```

With this pull request:

```
>>> import pexpect
>>> pexpect.run("a-script.sh", env={"PATH": "/tmp"})
'Running script not on $PATH\r\n'
```

bz2:
Is there any prospect of this getting landed? We had to workaround the bug downstream:

<http://bazaar.launchpad.net/~juju-qa/juju-ci-tools/trunk/revision/1409>

I think the prospect of subtle breakage from different PATH contents across machines is far more of a concern than the minor behaviour change given the current logic is inconsistent.

takluyver:
I still see it as a design choice rather than a bug, but I think consistency with subprocess is a pretty good argument for choosing the other option. If @jquast agrees, I think this can be merged. Thanks for adding tests.

takluyver:
@jquast if I don't hear from you in a couple of days, I'll assume you're happy with this.

jquast:
Reviewing now!

jquast:
This is perfectly reasonable. @bz2 is correct, this is precisely how Popen behaves, so it makes sense to mirror its behavior. I will merge, and add another commit to bump the version for next release and add a note to doc/history.rst.

I feel strongly that it should be noted about in doc/history.rst: Someone in the field may rely on the current processes PATH to find the target for execution while supplying a custom PATH that the target would not otherwise be found in.

My only review note: Looking at ``os.subprocess.Popen``, we could probably benefit from using ``os.get_exec_path`` directly for the same purposes, https://github.com/python/cpython/blob/bdf5dd57f19dcbf7d6ca00993cdfb6f39944384d/Lib/os.py#L690

Though i compared both versions a number of times, they are more or less identical, I think I'd rather favor augmenting which() as @bz2 has done, so OK as-is, and thank you very much for the tests!

jquast:
@takluyver Added commit 38ed591f830d575ac83640549820ce43c15ccb88 to bump to 4.2.0 and describe change, if you approve of it, go ahead and publish to pypi or revise at your discretion. I'm finding the tags very useful, I prepared a *draft* "release" tag at https://github.com/pexpect/pexpect/releases/edit/untagged-b30604874958c8feed1f

bz2:
Thanks guys!

PR #349, opened 2016-05-19
add serial spawn subclass

nodish:
this enable controlling serial port on windows.


takluyver:
Thanks for this! However, as I described on #340, I'm wary of adding more features to Pexpect, because I'd rather not spend a lot of time maintaining it. Ping @jquast - how do you feel about adding this?

jquast:
Good work @nodish, I really appreciate your contribution! Regretfully, I do not wish to add such feature to pexpect, as it means @takluyver and I must test, support, and manage issues or changes for this file in the future.

As pexpect does not include or require on pyserial, this should belong in an independent package. I suggest publishing a "pexpect-pyserial" or similar package on pypi. Best wishes!

PR #359, opened 2016-06-28
Restore POSIX support to fdpexpect.

MountainRider:
Factor out uninterruptible select and put the code in utils as `select_ignore_interrupts`. Modify `fdpexpect` to use `select_ignore_interrupts` if `os.name` is `posix`. This should restore POSIX support without affecting Windows support.

### DIFF
--- a/pexpect/fdpexpect.py
+++ b/pexpect/fdpexpect.py
@@ -115,10 +115,18 @@ class fdspawn(SpawnBase):
             self.write(s)
 
     def read_nonblocking(self, size=1, timeout=-1):
-        """The read_nonblocking method of SpawnBase assumes that a call to
+        """ Read from the file descriptor and return the result as a string.
+
+        The read_nonblocking method of SpawnBase assumes that a call to
         os.read will not block. This is not the case for POSIX file like
         objects like sockets and serial ports. So we use select to implement
-        the timeout on POSIX."""
+        the timeout on POSIX.
+
+        :param size:    Read at most size bytes
+        :param timeout: Wait timeout seconds for file descriptor to be ready
+                        to read. If -1, use self.timeout. If 0, poll.
+        :return:        String containing the bytes read
+        """
         if os.name == 'posix':
             if timeout == -1:
                 timeout = self.timeout
--- a/pexpect/fdpexpect.py
+++ b/pexpect/fdpexpect.py
@@ -22,7 +22,8 @@ PEXPECT LICENSE
 '''
 
 from .spawnbase import SpawnBase
-from .exceptions import ExceptionPexpect
+from .exceptions import ExceptionPexpect, TIMEOUT
+from .utils import select_ignore_interrupts
 import os
 
 __all__ = ['fdspawn']
@@ -112,3 +113,19 @@ class fdspawn(SpawnBase):
         "Call self.write() for each item in sequence"
         for s in sequence:
             self.write(s)
+
+    def read_nonblocking(self, size=1, timeout=-1):
+        '''The read_nonblocking method of fdspawn assumes that the file
+        will never block. This is not the case for all file like objects
+        that support fileno (e.g. POSIX sockets and serial ports). So we
+        use select to implement the timeout on POSIX.'''
+        if os.name == 'posix':
+            if timeout == -1:
+                timeout = self.timeout
+            rlist = [self.child_fd]
+            wlist = []
+            xlist = []
+            rlist, wlist, xlist = select_ignore_interrupts(rlist, wlist, xlist, timeout)
+            if self.child_fd not in rlist:
+                raise TIMEOUT('Timeout exceeded.')
+        return super(fdspawn, self).read_nonblocking(size)
--- a/pexpect/fdpexpect.py
+++ b/pexpect/fdpexpect.py
@@ -115,10 +115,10 @@ class fdspawn(SpawnBase):
             self.write(s)
 
     def read_nonblocking(self, size=1, timeout=-1):
-        '''The read_nonblocking method of fdspawn assumes that the file
-        will never block. This is not the case for all file like objects
-        that support fileno (e.g. POSIX sockets and serial ports). So we
-        use select to implement the timeout on POSIX.'''
+        """The read_nonblocking method of SpawnBase assumes that a call to
+        os.read will not block. This is not the case for POSIX file like
+        objects like sockets and serial ports. So we use select to implement
+        the timeout on POSIX."""
         if os.name == 'posix':
             if timeout == -1:
                 timeout = self.timeout
--- a/pexpect/utils.py
+++ b/pexpect/utils.py
@@ -1,6 +1,9 @@
 import os
 import sys
 import stat
+import select
+import time
+import errno
 
 
 def is_executable_file(path):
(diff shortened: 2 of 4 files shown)


tests/test_socket.py
```diff
@@ -0,0 +1,150 @@
...
+                session.read_nonblocking(size=4096)
+                # This read should timeout
+                session.read_nonblocking(size=4096)
+            except pexpect.TIMEOUT:
+                result = 1
+            exit(result)
+        test_proc = multiprocessing.Process(target=socket_fn, args=(self,))
+        test_proc.daemon = True
+        test_proc.start()
+        time.sleep(5.0)
```
jquast: not a fan of 5.0 second sleep, I have an slightly similar test, where I use semaphores between child/parent stdout, with multiprocessing library we have more simple primitives like https://docs.python.org/2/library/multiprocessing.html#multiprocessing.Event

https://github.com/jquast/blessed/blob/master/blessed/tests/test_keyboard.py#L38-L120

tests/test_socket.py
```diff
@@ -0,0 +1,150 @@
...
+            try:
+                sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+                sock.connect(('rainmaker.wunderground.com', 23))
+                session = fdpexpect.fdspawn(sock, timeout=10)
+                # Get all data from server
+                session.read_nonblocking(size=4096)
+                # This read should timeout
+                session.read_nonblocking(size=4096)
+            except pexpect.TIMEOUT:
+                result = 1
```
jquast: suggest something more irregular -- all python exceptions in child process will exit with a value of '1', so testing for '1' is not definitive. If we had a connection timeout or any other error, we would have a "positive" result without exercise

``python -c 'raise RuntimeError("Some error")'; echo $?`` prints 1.

takluyver:
@MountainRider would you be interested in co-maintaining Pexpect? You seem to care about it and have a good handle on how it works. It's up to @jquast as well whether we add another co-maintainer, but I think it would be good to have another person on it.

pexpect/utils.py
```diff
@@ -110,3 +113,32 @@ def split_command_line(command_line):
...
+    (terminal resize). '''
+
+    # if select() is interrupted by a signal (errno==EINTR) then
+    # we loop back and enter the select() again.
+    if timeout is not None:
+        end_time = time.time() + timeout
+    while True:
+        try:
+            return select.select(iwtd, owtd, ewtd, timeout)
+        except select.error:
```
MountainRider: I didn't write the code, I just moved it:
https://github.com/pexpect/pexpect/blob/master/pexpect/pty_spawn.py#L805
It it has problems on Python3 they were there before I moved it.

tests/test_socket.py
MountainRider: Not a fan of global, but I can change the test to use an Event.

tests/test_socket.py
MountainRider: Are you saying you want to see ECONNREFUSED? Or did you pick 111 out of a hat?
`#define ECONNREFUSED    111     /* Connection refused */`
If you really want something unique, I recommend going above the 255 limit on exit codes:
http://linux.die.net/man/3/exit

MountainRider:
@takluyver: Yes, I would be interested. You folks have done a wonderful job of maintaining Pexpect, and I have been a free-rider user for a long time. It would be more than fair for me to donate some of my time to the project.

pexpect/utils.py
jquast: Yes, I am only just making note of it for ourselves. Better in writing than back of my mind!

takluyver:
Cool! @jquast, is that OK by you?

I'll ask for your real identity, because I'm not quite comfortable giving push access to an pseudonymous Github handle. It needn't be publicly visible if you prefer to go by the pseudonym: drop me an email (address on my profile) if you're more comfortable that way.

On Pexpect and all the shared-maintainership projects I work on, we try to work in pull requests for any non-trivial changes, giving each other a chance to review things. Pushing to master is OK for things like correcting simple typos.

tests/test_socket.py
jquast: It is chosen out of a hat, it has no meaning. All python exceptions exit value 1, so we want to ensure we are testing something else, '2' would be just as well.

Side note, cannot use exit codes over 255 though, its a strict byte limitation, it just rolls around:
```
$ python -c 'exit(299)'; echo $?
43
```

jquast:
@takluyver I approve, agree also with everything @takluyver says.

Keep in mind that the project is in maintenance mode, only making new releases for bugfixes or performance. When we do our job well, we're making very few and very careful releases! Thanks for joining

MountainRider:
My name is David Cullen. I work for Technicolor in Lawrenceville, GA, USA. You can reach me at david.cullen@technicolor.com. If you want to talk on the phone, we can exchange numbers and I can arrange to pay the bill.

takluyver:
Thanks, that's good enough for me :-) You should be getting an invitation to join the team.

MountainRider:
Something seems broken with my pull request. The Python3 tests are failing on Travis CI, but they pass when I run them locally.

takluyver:
Certain operations are unusually slow on Travis, so it can expose race conditions that it's very hard to hit normally.

takluyver:
I see the tests are passing now. :-)

MountainRider:
ok. Everything finally passed. However, I would like some help getting coverage on this line:
https://coveralls.io/builds/6801173/source?filename=pexpect%2Futils.py#L141
I can't see the problem with the test. Any tips on how I can reach that line?

takluyver:
I think that line can only be reached with a race condition - the select would have to be interrupted just before it times out, so that by the time `time.time()` is called a few lines later, it's after `end_time`. So I wouldn't worry about not covering it.

MountainRider:
ok. I rigged up a harness to try to force the race condition, but I was unsuccessful. I guess @takluyver and @jquast need to review my latest changes.

pexpect/utils.py
MountainRider: ok. I followed the same pattern in the links you gave. We should now have Python3 coverage.

pexpect/fdpexpect.py
```diff
@@ -112,3 +113,19 @@ def writelines(self, sequence):
         "Call self.write() for each item in sequence"
         for s in sequence:
             self.write(s)
+
+    def read_nonblocking(self, size=1, timeout=-1):
+        '''The read_nonblocking method of fdspawn assumes that the file
+        will never block. This is not the case for all file like objects
```
takluyver: This docstring is a bit confusing - it would assume that if we weren't overriding it, but we are (on POSIX).

takluyver:
This looks OK to me, though I haven't reviewed the new tests.

pexpect/fdpexpect.py
MountainRider: Will this change fix the confusion:
`'''The read_nonblocking method of SpawnBase assumes that the file`

takluyver: It's better, but ideally the docstring should start by describing what this method does, rather than the method on the parent class. I like docstrings to follow the format:

```
One line overview of functionality
<blank line>
If necessary, more detailed description of options, return value, and rationale.
```

MountainRider:
This line look unreachable. Is this a copy-paste error?
https://github.com/pexpect/pexpect/blob/master/pexpect/pty_spawn.py#L406

takluyver:
It does. I guess I overlooked it when separating ptyprocess out of pexpect. Feel free to remove it, since it clearly can't be reached.

MountainRider:
ok. What next?

MountainRider:
ok. I saw some more docstring feedback from you when I checked my email. I have updated the docstring for the `read_nonblocking` method of `fdspawn`. Hopefully, it meets or exceeds your expectations.

takluyver:
Thanks, this is looking good. I'll give @jquast a chance to have another look, and then I think it can be merged.

jquast:
Looks great.

I'll add a changelog entry and prepare it for a version bump before merge to master in a bit after dinner. 

Thanks again @MountainRider we appreciate the help!

PR #360, opened 2016-06-30
format license according to the standard template

waldyrious:
Having "PEXPECT LICENSE" at the top of the file makes it seem like it is a custom license with specific terms, rather than the license of the Pexpect project. The text corresponds (as the header note righfully indicates) to the ISC License, so this change edits the header to hopefully clarify this.

Additionally, the formatting was slightly different from the standard ISC license template (as seen on https://opensource.org/licenses/isc-license or http://choosealicense.com/licenses/isc/), so the first paragraph was converted to regular sentence case rather than ALL CAPS like the disclaimer.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-PEXPECT LICENSE
+ISC LICENSE
 
     This license is approved by the OSI and FSF as GPL-compatible.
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks

PR #361, opened 2016-07-02
Prepare 4.2 release

jquast:
- bump version in documentation via pexpect/__init__.py
- describe PR #359 in changelog as regression
- minor docstring enhancements to fdspawn.read_nonblocking()


takluyver:
We haven't actually done a 4.2 release, so we might want to call this 4.2.

jquast:
Right, I noticed that when drafted the tag, will squash

jquast:
Merged and tagged, need only to publish to pypi.

I'd be happy to do so, using command ``python setup.py sdist bdist_wheel upload`` with your ok @takluyver, or you can also do the same. Thanks!

jquast:
4.2.0 published

PR #362, opened 2016-07-09
Fix running 'env' in replwrap-ed bash

takluyver:
From takluyver/bash_kernel#33 - if you run `env` or `echo $PS1` inside a replwrap bash, the output will contain the prompt string. Pexpect will think that is the prompt, cut off the output there, and then get into a confused state about the output of future commands.

This avoids that by inserting some characters `\[\]` which are displayed when inspecting the environment variable, but invisible when bash actually renders the prompt. This means that the output of `env` or `echo $PS1` doesn't match the prompt replwrap is looking for.

Those characters are intended to mark a set of non-printing characters (e.g. to change the terminal colour), so having them together as a pair should have no ill effects. I've verified in interactive testing that it works.

### DIFF
--- a/pexpect/replwrap.py
+++ b/pexpect/replwrap.py
@@ -109,5 +109,14 @@ def bash(command="bash"):
     bashrc = os.path.join(os.path.dirname(__file__), 'bashrc.sh')
     child = pexpect.spawn(command, ['--rcfile', bashrc], echo=False,
                           encoding='utf-8')
-    return REPLWrapper(child, u'\$', u"PS1='{0}' PS2='{1}' PROMPT_COMMAND=''",
+
+    # If the user runs 'env', the value of PS1 will be in the output. To avoid
+    # replwrap seeing that as the next prompt, we'll embed the marker characters
+    # for invisible characters in the prompt; these show up when inspecting the
+    # environment variable, but not when bash displays the prompt.
+    ps1 = PEXPECT_PROMPT[:5] + u'\[\]' + PEXPECT_PROMPT[5:]
+    ps2 = PEXPECT_CONTINUATION_PROMPT[:5] + u'\[\]' + PEXPECT_CONTINUATION_PROMPT[5:]
+    prompt_change = u"PS1='{0}' PS2='{1}' PROMPT_COMMAND=''".format(ps1, ps2)
+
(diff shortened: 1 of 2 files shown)


pexpect/replwrap.py
```diff
@@ -109,5 +109,14 @@ def bash(command="bash"):
     bashrc = os.path.join(os.path.dirname(__file__), 'bashrc.sh')
     child = pexpect.spawn(command, ['--rcfile', bashrc], echo=False,
                           encoding='utf-8')
-    return REPLWrapper(child, u'\$', u"PS1='{0}' PS2='{1}' PROMPT_COMMAND=''",
+
+    # If the user runs 'env', the value of PS1 will be in the output. To avoid
+    # replwrap seeing that as the next prompt, we'll embed the marker characters
+    # for invisible characters in the prompt; these show up when inspecting the
+    # environment variable, but not when bash displays the prompt.
+    ps1 = PEXPECT_PROMPT[:5] + u'\[\]' + PEXPECT_PROMPT[5:]
```
MountainRider: What does the magic number 5 mean?

tests/test_replwrap.py
```diff
@@ -33,6 +33,13 @@ def test_pager_as_cat(self):
         res = bash.run_command('man sleep', timeout=5)
         assert 'SLEEP' in res, res
 
+    def test_bash_env(self):
```
MountainRider: Maybe add a comment explaining the goal of the test?

pexpect/replwrap.py
takluyver: It's an arbitrary position to insert the extra characters. It has to be neither at the beginning nor the end.

tests/test_replwrap.py
takluyver: Thanks, done.

jquast:
Very good, merging.

(Of course need to bump, changelog, and release for IPython to pick it up)

Reminds me of the old trick ``ps auxww | grep comman[d]`` -- ensures grep command does not match itself.

jquast:
You may have received an e-mail of a failed build from Travis. This message is erroneous, job succeeded when retried (python3.5 tests).

takluyver:
Thanks both :)

PR #376, opened 2016-09-07
dont remove reference to _UnixReadPipeTransport

gescheit:
_UnixReadPipeTransport class closes pipe in __del__ method. This commit
adds attribute to SpawnBase with pointer to current transport.

### DIFF
--- a/pexpect/async.py
+++ b/pexpect/async.py
@@ -12,22 +12,30 @@ def expect_async(expecter, timeout=None):
     idx = expecter.new_data(previously_read)
     if idx is not None:
         return idx
-
-    transport, pw = yield from asyncio.get_event_loop()\
-        .connect_read_pipe(lambda: PatternWaiter(expecter), expecter.spawn)
-
+    if not expecter.spawn.async_pw_transport:
+        pw = PatternWaiter()
+        pw.set_expecter(expecter)
+        transport, pw = yield from asyncio.get_event_loop()\
+            .connect_read_pipe(lambda: pw, expecter.spawn)
+        expecter.spawn.async_pw_transport = pw, transport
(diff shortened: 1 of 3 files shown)


pexpect/async.py
```diff
@@ -12,22 +12,30 @@ def expect_async(expecter, timeout=None):
...
     except asyncio.TimeoutError as e:
         transport.pause_reading()
         return expecter.timeout(e)
 
+
 class PatternWaiter(asyncio.Protocol):
     transport = None
-    def __init__(self, expecter):
+
+    def set_expecter(self, expecter):
```
takluyver: I don't really like this pattern of having a non-constructor method which must be called before other methods will work.

I've probably used asyncio wrong in some way when writing this code, but I don't really have the time or interest to work out the right way to do it at the moment.

gescheit: I think that you can apply this workaround with "no-like" pattern and resolve critical bug. When you got more free time you can make more accurate resolution of this problem. At this time expect() with "async" option totally unusable. Also need to rename async because in python 3.5 async is name of operator.

MountainRider: Why can't the expecter be passed to the constructor? Was the constructor removed so that the super class constructor would be called instead? I don't understand this change.

gescheit: Because instance of PatternWaiter should not be destroyed. If this happens socket will be closed. So i save PatternWaiter instance in SpawnBase instance.

mbrar:
Hi there, I would really like to see this code merged in. Thanks!

takluyver:
OK, there's little point in holding out for a better solution if I'm not prepared to work on one.

Issue #383, opened 2016-10-17
ptyprocess.PtyProcessError raised on close


kristopher-h:
I ran into a situation where a ptyprocess.PtyProcessError is raised when trying to close a pty_spawn.spawn object and that is not possible (ptyprocess.ptyprocess.PtyProcessError: Could not terminate the child.). I'm thinking the call to self.ptyproc.close should be wrapped in _wrap_ptyprocess_err? This to avoid having to be aware of ptyprocess exceptions in higher layers.

E.g.:

```
def close(self, force=True):
    '''This closes the connection with the child application. Note that
    calling close() more than once is valid. This emulates standard Python
    behavior with files. Set force to True if you want to make sure that
    the child is terminated (SIGKILL is sent if the child ignores SIGHUP
    and SIGINT). '''

    self.flush()
    with _wrap_ptyprocess_err():
        # PtyprocessError can be raised if it is not possible to close child process.
        self.ptyproc.close(force=force)
    ...
```

takluyver:
Sounds reasonable. Want to make a PR?

kristopher-h:
Sure thing I can give it a go (will probably take me a couple of days to figure all the ins and outs of github out, setting up the env etc, but that could be fun).

takluyver:
Great :-). Let me know if you run into any problems.

takluyver:
Fixed by #386

PR #386, opened 2016-10-19
Raise ExceptionPexpect instead of PtyProcessError in spawn.close

kristopher-h:
Pull request to solve issue described in https://github.com/pexpect/pexpect/issues/383.

//Kristopher


takluyver:
Thanks! This looks OK to me, I'll give it a while for @jquast and @MountainRider to have a look if they want to.

PR #389, opened 2016-10-31
Remove reference to missing script

lazymutt:
Remove reference to missing sshls.py script in examples


takluyver:
Thanks, and sorry for the delayed response. Looks like it's not the only one listed there that's no longer present.

PR #394, opened 2016-12-04
Pxssh refactor

Flexin1981:
Hi Guys,

Please feel free to ignore this, but I use this module on a daily basis and thought that it may be worthy of a clean up. There is no documentation on how you would like pull requests to be initiated, so forgive me for any lapses in etiquette.

Any questions or changes that you would like or anything that you think that I have missed let me know

### DIFF
--- a/pexpect/pxssh.py
+++ b/pexpect/pxssh.py
@@ -61,7 +61,7 @@ class pxssh(spawn):
         """
         options = ''.join([" -o '%s=%s'" % (o, v) for (o, v) in self.options.items()])
 
-        for key, val in pxssh._SSH_OPTIONS.iteritems():
+        for key, val in pxssh._SSH_OPTIONS.items():
             try:
                 options += val.format(getattr(self, key))
             except AttributeError:
--- a/pexpect/pxssh.py
+++ b/pexpect/pxssh.py
@@ -246,6 +246,10 @@ class pxssh(spawn):
                                          self.before, self.PROMPT,))
 
     def logout(self):
+        """
+            Sends exit to the remote shell.
+            If there are stopped jobs then this automatically sends exit twice.
+        """
 
         self.sendline("exit")
         index = self.expect([EOF, "(?i)there are stopped jobs"])
@@ -274,6 +278,21 @@ class pxssh(spawn):
 
     def set_unique_prompt(self):
 
+        """
+            This sets the remote prompt to something more unique than ``#`` or ``$``.
+            This makes it easier for the :meth:`prompt` method to match the shell prompt
+            unambiguously. This method is called automatically by the :meth:`login`
+            method, but you may want to call it manually if you somehow reset the
+            shell prompt. For example, if you 'su' to a different user then you
+            will need to manually reset the prompt. This sends shell commands to
+            the remote host to set the prompt, so this assumes the remote host is
+            ready to receive commands.
+            Alternatively, you may use your own prompt pattern. In this case you
+            should call :meth:`login` with ``auto_prompt_reset=False``; then set the
+            :attr:`PROMPT` attribute to a regular expression. After that, the
+            :meth:`prompt` method will try to match your prompt pattern.
+        """
+
         self.sendline("unset PROMPT_COMMAND")
         self.sendline(self.PROMPT_SET_SH)  # sh-style
         i = self.expect([TIMEOUT, self.PROMPT], timeout=10)
@@ -285,6 +304,16 @@ class pxssh(spawn):
         return True
 
     def sync_original_prompt(self):
+
+        """
+            This attempts to find the prompt. Basically, press enter and record
+            the response; press enter again and record the response; if the two
+            responses are similar then assume we are at the original prompt.
+            This can be a slow function. Worst case with the default sync_multiplier
+            can take 12 seconds. Low latency connections are more likely to fail
+            with a low sync_multiplier. Best case sync time gets worse with a
+            high sync multiplier (500 ms with default).
+        """
 
         self.sendline('')
         x = self.read_prompt()
--- a/pexpect/pxssh.py
+++ b/pexpect/pxssh.py
@@ -1,4 +1,5 @@
-'''This class extends pexpect.spawn to specialize setting up SSH connections.
+"""
+This class extends pexpect.spawn to specialize setting up SSH connections.
 This adds methods for login, logout, and expecting the shell prompt.
 
 PEXPECT LICENSE
@@ -17,12 +18,10 @@ PEXPECT LICENSE
     WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
     ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
     OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+"""
 
-'''
 
 from pexpect import ExceptionPexpect, TIMEOUT, EOF, spawn
-from pprint import pprint
-import copy
 
 
 __all__ = ['ExceptionPxssh', 'pxssh']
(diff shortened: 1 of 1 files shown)


takluyver:
Hi, and thanks! We need to be quite careful not to break stuff, as Pexpect is mature code, but so long as we can do that,  refactoring is OK.

There seem to be some test failures at the moment - can you have a look at them?

Flexin1981:
Hey guys,

I have fixed the small issues and the tests pass on my machine, however the test run seems to fail on here, I am not sure that it is the tests for the pxssh module that are failing, I am not familiar with the testing output on the travis tool, can someone confirm this for me ?

Also with regards to the testing suite for this I am not sure that the testing is sufficient to ensure that the code is working, is it ok to expand this area too ?

pexpect/pxssh.py
```diff
@@ -21,389 +21,234 @@
...
+    UNIQUE_PROMPT = "\[PEXPECT\][\$\#] "
+
+    def _build_ssh_command(self):
+        """
+            Method to build the ssh command string.
+        :return:
+        """
+        options = ''.join([" -o '%s=%s'" % (o, v) for (o, v) in self.options.items()])
+
+        for key, val in pxssh._SSH_OPTIONS.iteritems():
```
takluyver: `.iteritems()` is gone in Python 3 (`.items()` returns an iterator anyway). It's possible to use a wrapper function that calls `.iteritems()` on Py2, but as there's only 4 things in the dict anyway, let's just use `.items()` on it.

takluyver:
There's one failure that's really still associated with this, but there is also a problem with Travis; I'll try to fix that on master now.

Flexin1981:
I have fixed the iteritems issue

takluyver:
Thanks, it looks like it's working. I'll try to have a closer look at the changes soon.

ofek:
@Flexin1981 Mind rebasing?

Flexin1981:
@ofek of course not, that is done and all the tests appear to be passing.

ofek:
@Flexin1981 Yay, thanks!

@takluyver This appears ready for merge.

pexpect/pxssh.py
```diff
@@ -21,389 +21,234 @@
...
-        # disable only SSH_ASKPASS without also disabling X11 forwarding.
-        # Unsetting SSH_ASKPASS on the remote side doesn't disable it! Annoying!
-        #self.SSH_OPTS = "-x -o'RSAAuthentication=no' -o 'PubkeyAuthentication=no'"
-        self.force_password = False
+class ExceptionPxssh(ExceptionPexpect):
+    pass
+
+
+class pxssh(spawn):
+
```
takluyver: You seems to have got rid of all docstrings? Why would you do that?

Flexin1981: hey, the code seemed relatively simple in nature and I have always  be taught to not pollute the code, so I guess  i am saying this came down to style.

If this is the roadblock I can add some relevant docstrings back in, not a problem

ofek: @Flexin1981 Yes please, just put whatever documentation that was there previously back in.

Flexin1981: Ok guys, I have added a lot of the code commenting back in, that I feel comfortable that is still applicable, I agree that perhaps this can be augmented by some new comments along the way.

Let  me know what you think.

pexpect/pxssh.py
```diff
@@ -17,393 +18,305 @@
...
+        if self.auto_prompt_reset:
             if not self.set_unique_prompt():
                 self.close()
                 raise ExceptionPxssh('could not set shell prompt '
                                      '(received: %r, expected: %r).' % (
                                          self.before, self.PROMPT,))
-        return True
 
-    def logout (self):
-        '''Sends exit to the remote shell.
```
takluyver: This one has still gone.

pexpect/pxssh.py
```diff
@@ -17,393 +18,305 @@
...
-        shell prompt. For example, if you 'su' to a different user then you
-        will need to manually reset the prompt. This sends shell commands to
-        the remote host to set the prompt, so this assumes the remote host is
-        ready to receive commands.
-
-        Alternatively, you may use your own prompt pattern. In this case you
-        should call :meth:`login` with ``auto_prompt_reset=False``; then set the
-        :attr:`PROMPT` attribute to a regular expression. After that, the
-        :meth:`prompt` method will try to match your prompt pattern.
-        '''
```
takluyver: And this

pexpect/pxssh.py
```diff
@@ -17,393 +18,305 @@
...
+        if i == 0:  # csh-style
             self.sendline(self.PROMPT_SET_CSH)
             i = self.expect([TIMEOUT, self.PROMPT], timeout=10)
             if i == 0:
                 return False
         return True
 
-# vi:ts=4:sw=4:expandtab:ft=python:
+    def sync_original_prompt(self):
+
```
takluyver: And this one had a docstring too

pexpect/pxssh.py
takluyver: Thanks! Docs and comments are not pollution, they're valuable hints to people who want to use it or modify it.

pexpect/pxssh.py
Flexin1981: added

pexpect/pxssh.py
Flexin1981: added

pexpect/pxssh.py
Flexin1981: added

takluyver:
Checking through the public methods, it looks like `levenshtein distance` and `try_read_prompt` have gone. I suspect the first of those is unlikely to be used externally, but the second one might be.

I'm a bit wary about this because it's a big change that I don't really have time to review in proper detail - I'm just skimming though picking up a few things that look wrong to me. I know from my own attempts to refactor pexpect that people can and do rely on details that didn't seem important to me.

@ofek have you used this branch, or looked through the code in detail?

Flexin1981:
@takluyver 

I agree that this is a big change to the code, but i think in a way that could be of benefit, perhaps we go about this kind of a refactor in a different way.

My thought is to expand the tests to cover more of the code lines, and what the functionality should be, that way you would be more confident that the changes won't change the behavior. we could also refactor the code in a more transformative way and not such a large chunk.

Let me know your thoughts

ofek:
@takluyver Nope. Honestly I just noticed a seemingly complete PR that needed a bump.

MountainRider:
I am against merging this because

- None of the maintainers has closely reviewed the changes
- It is a big change that does not fix any defects
- It might break things that are currently working
- Comments were removed for no good reason

These are all signs of a naive "code cleanup". I have seen this type of thing before and have even perpetrated this type of thing myself. Changing working code inevitably introduces defects. These defects do not always show up immediately. Do we really want to break someone's working code?

pexpect/pxssh.py
```diff
@@ -17,393 +18,334 @@
     WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
     ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
     OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+"""
 
-'''
```
MountainRider: Changes like this add no value.

Flexin1981:
Hey Guys,

Like I said in my first comment, if this is not for you then feel free to close this off, my intention here has always been to improve what is there.

The fact that we are picking on the comments as the reason not to do this is seems to be skipping the larger point that you don't trust your tests to encompass all your behavior and hence you have become coupled to the code that you have.

If there is anything to take away from this, please recognize the need to expand your testing so that changes to the code can happen freely without fear of change and so that outside developers can "pitch in" without having to have a intimate knowledge of the intricacies of the behavior. 

Again as i have said above, if this is something that you guys would like to change I am happy to put in some time trying to pull out all the behaviors that you have into tests so that over time you can be more comfortable with this kind of change.

takluyver:
OK, I think we're going to close this. Sorry that we left it this long and asked you to make more changes rather than closing it sooner.

Pexpect is old code, and like many Python modules, no-one ever really defined the Public API that downstream developers can rely on. That means that code using Pexpect relies on features that seem incidental to us, and people come and file bugs when we change them. You're right that the tests don't encompass all the behaviour we need to preserve, and they probably never will.

I'm not sure about the other maintainers, but I'm now happy for Pexpect to get only occasional bugfixes and updates to keep it working with new versions of Python. If people want major new features or refactoring the code into a cleaner form, I'd recommend that they fork it under a new name - or understand what it does and start a new version from scratch. [Streamexpect](https://github.com/digidotcom/python-streamexpect) was one such effort, but it doesn't look like it's actively developed any more.

PR #395, opened 2016-12-04
(minor) Remove spawnu in spawn docstring in favour of spawn with encoding arg

AngusP:
Update docstring for class `spawn.__init__` in pty_spawn.py to remove reference to deprecated spawnu in favour of spawn with ``encoding`` keyword argument

``spawnu`` deprecated according to line 805 of `pty_spawn.py` and issue #159

### DIFF
--- a/pexpect/pty_spawn.py
+++ b/pexpect/pty_spawn.py
@@ -106,8 +106,9 @@ class spawn(SpawnBase):
             child = pexpect.spawn('some_command')
             child.logfile = sys.stdout
 
-            # In Python 3, spawnu should be used to give str to stdout:
-            child = pexpect.spawnu('some_command')
+            # In Python 3, spawn with the argument ``encoding`` should be used to ensure utf-8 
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks, this is a good catch. I'm going to quibble over the wording a bit, however:

- We can't ensure that the data the process sends is in any particular encoding - that's up to the subprocess. What the encoding parameter affects is how we encode data that we send to the pty (i.e. stdin of the subprocess), and how we *try* to decode data coming back. As a special case, when `encoding=None`, which is the default, the interface uses bytes rather than text.
- I tend to avoid talking about stdout or stderr when running subprocesses in a pty, because we have just the one pty that combines both of them (and in fact stdin as well, but that's a bit different).

AngusP:
@takluyver feel free to adjust the wording, I've only recently found this library so haven't familiarised with its internals.

takluyver:
OK, thanks for spotting it, in any case.

PR #417, opened 2017-05-25
added missing assignment of pid var to PopenSpawn class

TheNuclearCat:
The pid of the class PopenSpawn has always returned None because it was never assigned


jquast:
Thanks!!

PR #430, opened 2017-07-17
Use raw strings for strings with literal backslash

nresare:
In python 3.6 it raises a DeprecationWarning to have strings with invalid backslash escape sequences. Changing the strings to raw strings (that doesn't try to look for escape sequences) makes the intention clearer and shuts up the DeprecationWarning


takluyver:
Thanks :-)

PR #445, opened 2017-09-15
Python 3.6 invalid escape sequence deprecation fixes

scop:
https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior

### DIFF
--- a/pexpect/replwrap.py
+++ b/pexpect/replwrap.py
@@ -114,9 +114,9 @@ def bash(command="bash"):
     # replwrap seeing that as the next prompt, we'll embed the marker characters
     # for invisible characters in the prompt; these show up when inspecting the
     # environment variable, but not when bash displays the prompt.
-    ps1 = PEXPECT_PROMPT[:5] + u'\[\]' + PEXPECT_PROMPT[5:]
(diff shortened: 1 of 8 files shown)


pexpect/replwrap.py
```diff
@@ -114,9 +114,9 @@ def bash(command="bash"):
     # replwrap seeing that as the next prompt, we'll embed the marker characters
     # for invisible characters in the prompt; these show up when inspecting the
     # environment variable, but not when bash displays the prompt.
-    ps1 = PEXPECT_PROMPT[:5] + u'\[\]' + PEXPECT_PROMPT[5:]
-    ps2 = PEXPECT_CONTINUATION_PROMPT[:5] + u'\[\]' + PEXPECT_CONTINUATION_PROMPT[5:]
+    ps1 = PEXPECT_PROMPT[:5] + ur'\[\]' + PEXPECT_PROMPT[5:]
```
takluyver: Looks like this is not valid syntax on Python 3. I think you'll have to double up the slashes here instead: `u'\\[\\]'`.

scop: Oops, fixed.

takluyver:
Thanks!

PR #446, opened 2017-09-15
Fix popen_spawn when specifying a path on Windows which contains "\" …

eli-b:
…separators


techalchemy:
As I mentioned on the referenced PR, it might be a better approach to use `posix=(os.name == 'posix')` when testing for posix compliance...

eli-b:
I've updated the PR according to your suggestion.

takluyver:
Which referenced PR are you discussing?

eli-b:
I think @techalchemy meant [this PR](https://github.com/kennethreitz/pipenv/pull/564).

takluyver:
Thanks, I should have seen that.

PR #447, opened 2017-09-27
Don't split cmd strings on win32 in PopenSpawn

rob356:
[On win32 Popen uses strings for args](https://docs.python.org/3/library/subprocess.html#converting-argument-sequence), so in PopenSpawn it isn't necessary to convert a string to a list just for Popen to convert it back again.
This also resolves an issue I was having where if the path to the executable was double quoted those quotes would stay in the first element of the cmd list and would cause Popen to throw a FileNotFoundError.


takluyver:
Seems reasonable, thanks.

blink1073:
@takluyver, can I pester you for a patch release with this?

PR #449, opened 2017-10-14
Updates README.rst with svg badge

sobolevn:


takluyver:
Is there a problem with using png? I think svg support is still somewhat patchy across different browsers, so I'd rather stick with what I know works.

sobolevn:
Thanks for your feedback!
All major browsers support `svg`.
Proof: https://caniuse.com/#search=svg

Moreover, developers use new versions.
`png` works, but it looks terrible on retina. 

Proof:


Compared to `svg`:

takluyver:
I think they all support 'SVG', but not necessarily all parts of SVG. Presumably Travis' badge uses things that are likely to work across all browsers, though. Let's go for this.

Issue #315, opened 2016-01-08
Issues with asyncio on Python 3.5


takeda:
It's very difficult to use methods that have async parameter such as expect() in Python 3.5.

The keyword 'async' is reserved in async coroutines (e.g. "async def function") which causes python to throw syntax error.

If possible instead of renaming the variable can pexpect support asyncio in similar fashion to for example https://github.com/cablehead/python-consul

for example

```
c = consul.Consul()
```

uses blocking mode

```
c = consul.aio.Consul()
```

uses asyncio

Passing `async = true` each time when using expect() and related methods is a bit annoying.

jquast:
right, our ``async`` keyword change predated that PEP's acceptance, sad.

takluyver:
It's not quite that bad - there was some trickery done with the syntax so that async is not (yet?) a full keyword, and it's not a syntax error to use it as a name:

```
Python 3.5.1 |Continuum Analytics, Inc.| (default, Dec  7 2015, 11:16:01) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> def foo(async=False):
...   pass
... 
>>> foo(async=True)
```

But I agree that it's not ideal, and we would have picked a different name if we were designing it after the new syntax landed for Python 3.5.

I'd rather not have a separate async class, because there's already a class hierarchy for different situations (fdspawn, pxssh, etc.), and I don't want to have to have parallel sync/async versions of each - that's the same mess we backed away from with having bytes/unicode as separate classes.

What about allowing it as a constructor parameter, so you would:

```python
c = pexpect.spawn(['blah'], asyncio=True)
i = yield from c.expect()
```

takeda:
It actually is bad, async is enforced inside of async function e.g.:

```python
#! /usr/bin/env python3.5

import asyncio
from contextlib import closing

import pexpect

async def main():
        p = pexpect.spawnu('uptime')
        await p.expect(r'up\s+(.*?),\s+([0-9]+) users?,\s+load averages?: ([0-9]+\.[0-9][0-9]),?\s+([0-9]+\.[0-9][0-9]),?\s+([0-9]+\.[0-9][0-9])', async = True)
        duration, users, av1, av5, av15 = p.match.groups()

        print(duration, users, av1, av5, av15)

if __name__ == '__main__':
        with closing(asyncio.get_event_loop()) as loop:
                loop.run_until_complete(main())
```
[...]

PR #450, opened 2017-10-21
Fix 37 compat

tacaswell:
Same idea as https://github.com/pexpect/pexpect/issues/315


tacaswell:
https://www.python.org/dev/peps/pep-0492/#deprecation-plans for the relevant PEP

takluyver:
Makes sense, thanks.

Issue #451, opened 2017-10-25
pxssh fails to recognize password prompt when banner contains the word 'password'


orulz:
On eg. line 280 of pxssh.py, quoted below:
`i = self.expect(["(?i)are you sure you want to continue connecting", original_prompt        , "(?i)(?:password:)|(?:passphrase for key)", "(?i)permission denied", "(?i)terminal type",         TIMEOUT, "(?i)connection closed by remote host", EOF], timeout=login_timeout)`
You can see it is looking for the word 'password' to signal that the password prompt has been received from the spawned ssh client.

However, if the device in question has the word 'password' in its banner, this logic fails and pxssh can't log in.

It seems that by changing the check from 'password' to 'password:' (on lines 280, 288, 298, 294) this would fixes the problem, and my experimentation verifies that this does seem to be the case and also doesn't seem to have any negative consequences (is there any instance when ssh would prompt for a password without a colon?) but I'm posting this here to get feedback in case I've missed something.

orulz:
I believe fixing this issue will also address issue #353

mfrlin:
Btw this fix breaks some Extreme routers that have prompt in style: `Enter password for <username>:`

We solved the issue by subclassing pxssh class and changing `login` method. However this requires a lot of copying and pasting code.

Here is a quick concept how login procedure could be extendable (untested): https://gist.github.com/mfrlin/52bc3117cfb9ad5634c012fae472b457

MountainRider:
I made this suggestion when we were reviewing the pull request:

> Wouldn't a more general solution be to add a method that allows the user to change the string used to expect the password prompt? This would avoid the possibility that changing the default would break any existing code.

Now we have an example of this change breaking existing code. Maybe this idea looks better now.

Red-M:
Hopefully my patch should fix this issue and another that I've spotted in the wild.

Red-M:
Does anyone encounter this issue after using the new password option for logging in?

mfrlin:
I did not test it. After using pexpect for a while with a lot of different systems I came to realization that login method is never going to be perfect. I think it is good enough as is as it handles most cases.

What would be good however is to refactor code, so you could write your own login procedure. As it is now subclassing pxssh and overriding login method feels kinda wonky. I don't have the fix or idea how to fix it yet, just saying that default login procedure is never going to be perfect for everyone.

Red-M:
The option should make this a lot more robust if you know the MOTD to begin with.

Issue #353, opened 2016-06-09
login method in pxssh is throwing exception password refused for correct password but going expire in near


Arunakumari1989:
In pxssh.py , login method because of the "(?:password") in expects the warning message "Warning: your password will expire in 3 days" ,  treating it as password prompt, so finally it is coming out with password refused exception for the correct password which is going to expire in near.

i = self.expect(["(?i)are you sure you want to continue connecting", original_prompt, "(?i)(?:password)|(?:passphrase for key)", "(?i)permission denied", "(?i)terminal type", TIMEOUT, "(?i)connection closed by remote host"], timeout=login_timeout)

        # First phase
        if i==0:
            # New certificate -- always accept it.
            # This is what you get if SSH does not have the remote host's
            # public key stored in the 'known_hosts' cache.
            self.sendline("yes")
            i = self.expect(["(?i)are you sure you want to continue connecting", original_prompt, "(?i)(?:password)|(?:passphrase for key)", "(?i)permission denied", "(?i)terminal type", TIMEOUT])
        if i==2: # password or passphrase
            self.sendline(password)
            i = self.expect(["(?i)are you sure you want to continue connecting", original_prompt, "(?i)(?:password)|(?:passphrase for key)", "(?i)permission denied", "(?i)terminal type", TIMEOUT])
        if i==4:
            self.sendline(terminal_type)
            i = self.expect(["(?i)are you sure you want to continue connecting", original_prompt, "(?i)(?:password)|(?:passphrase for key)", "(?i)permission denied", "(?i)terminal type", TIMEOUT])

        # Second phase
        if i==0:
            # This is weird. This should not happen twice in a row.
            self.close()
            raise ExceptionPxssh('Weird error. Got "are you sure" prompt twice.')
        elif i==1: # can occur if you have a public key pair set to authenticate.
            ### TODO: May NOT be OK if expect() got tricked and matched a false prompt.
            pass
        elif i==2: # password prompt again
            # For incorrect passwords, some ssh servers will
            # ask for the password again, others return 'denied' right away.
            # If we get the password prompt again then this means
            # we didn't get the password right the first time.
            self.close()
            raise ExceptionPxssh('password refused')

PR #452, opened 2017-10-25
Issue #451: Check for "password:" as prompt instead of "password"

orulz:
pxssh checks for 'password' as the password prompt rather than 'password:'. This breaks things when the word 'password' is in the banner - which is rather more common than you might think. I think what is happening is, pxssh is sending the password, but then seeing the word 'password' in the banner and thinking that it has failed to log in.

This extremely simple fix should work, as long as:

1. There aren't any legitimate instances where SSH will prompt for password without a colon (I have never observed this)
2. It still doesn't cover the case where the banner actually contains the word password followed by a colon - "password:" - handling this would require a more substantial change.

I would respectfully suggest consideration of merging this pull request unless there are legitimate known situations where the password prompt comes without a colon.

Any feedback on this?

### DIFF
--- a/pexpect/pxssh.py
+++ b/pexpect/pxssh.py
@@ -277,7 +277,7 @@ class pxssh (spawn):
         # This does not distinguish between a remote server 'password' prompt
         # and a local ssh 'passphrase' prompt (for unlocking a private key).
         spawn._spawn(self, cmd)
-        i = self.expect(["(?i)are you sure you want to continue connecting", original_prompt, "(?i)(?:password)|(?:passphrase for key)", "(?i)permission denied", "(?i)terminal type", TIMEOUT, "(?i)connection closed by remote host", EOF], timeout=login_timeout)
+        i = self.expect(["(?i)are you sure you want to continue connecting", original_prompt, "(?i)(?:password:)|(?:passphrase for key)", "(?i)permission denied", "(?i)terminal type", TIMEOUT, "(?i)connection closed by remote host", EOF], timeout=login_timeout)
 
         # First phase
         if i==0:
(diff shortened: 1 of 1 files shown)


orulz:
I believe this pull request will address Issues #353 and #451

takluyver:
Seems fine to me. I'll give @jquast and @MountainRider a couple of days to comment in case they know of any cases where the colon isn't there. If you don't hear anything else in a few days, ping me and I'll merge it.

MountainRider:
Wouldn't a more general solution be to add a method that allows the user to change the string used to expect the password prompt? This would avoid the possibility that changing the default would break any existing code.

jnatschev:
Not being a guru at pexpect (pxssh), couldn’t the ssh command include “-q”? You wouldn’t then need to worry about the string “password” as part of any other text, except for the actual “password” prompt.



Cheers,
John Natschev
‌

MountainRider:
Does `-q` suppress banners?

` -q      Quiet mode.  Causes most warning and diagnostic messages to be suppressed.`

orulz:
@MountainRider The confusion happens when the banner is sent by the remote host AFTER a login. Login is successful but pxssh thinks that it isn't successful because the banner contains the word "password". Even with ssh -q, the remote host will still send its banner after login, so this would not solve the problem.

takluyver:
Configurability is fine, but it makes sense to have sensible defaults as well. This is a simple change, and I'm in favour of merging it unless we know of a situation that it will break.

jnatschev:
Oh. OK. That’s a different story. Now it’s the difference between “banner” text pre-login and message-of-the-day text post-login.

Also, that’s a bit weird. Once the “real” password prompt has been detected and the password is sent, the next is to expect a prompt and not password.



Cheers,
John Natschev
‌

PR #457, opened 2017-11-21
handle more OSes

zimbatm:
ubuntu: /etc/bash.bashrc
centos: /etc/bashrc
nixos: /etc/bashrc


takluyver:
Thanks, this looks good to me.

PR #459, opened 2017-12-08
Unset PROMPT_COMMAND in bash for replwrap

takluyver:
People may use PROMPT_COMMAND to modify the prompt (hat-tip to @Naereen, who does). Unset it to avoid breaking replwrap.

The line between customisations that people might want to be preserved in replwrap (like aliases) and those that don't make sense (like custom prompts) is somewhat indistinct. But from the uses of PROMPT_COMMAND I've seen, I think it's more likely that it shouldn't apply in replwrap.

### DIFF
--- a/pexpect/bashrc.sh
+++ b/pexpect/bashrc.sh
@@ -11,3 +11,6 @@ fi
 
 # Reset PS1 so pexpect can find it
 PS1="$"
+
+# Unset PROMPT_COMMAND, so that it can't change PS1 to something unexpected.
+unset PROMPT_COMMAND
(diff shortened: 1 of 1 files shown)


Naereen:
Great!

Naereen:
Great!
Do you know when this will be available through pip and in `bash_kernel` ?

takluyver:
When I do a pexpect release ;-). Is it something you'd still like?

Naereen:
Well my bug was resolved, but it would be interesting to try it again on my previous PS1/PS2 configuration when possible. (No hurry at all of course!)

takluyver:
OK, 4.3.1 is out. :-)

Issue #438, opened 2017-08-11
pexpect has very poor performance when searching large output streams 


ryanpetrello:
Even with `searchwindowsize` set, pexpect's read/search algorithm seems to have _very poor_ performance for processes that write large quantities of data to stdout or stderr.  It gets significantly worse as the size of the output stream moves upwards.

`cat expect.py`
```python
#!/usr/bin/env python
import subprocess
import expect
# Write a ~5MB file
subprocess.call('base64 /dev/urandom | head -c 5000000 > /tmp/large', shell=True)
child = pexpect.spawn(
    'cat', ['/tmp/large'], [], env={}, ignore_sighup=True,
    encoding='utf-8', echo=False
)
while child.isalive():
    child.expect([], timeout=10000000, searchwindowsize=100)
```

When `cat /tmp/large` is run outside of pexpect on moderate hardware, it finishes (and prints) in seconds:

```
$ time cat /tmp/large
...
real	0m1.011s
user	0m0.000s
sys	0m0.130s
```

When run with pexpect, it takes notably longer:

```
time python expect.py
real	0m5.451s
user	0m1.970s
sys	0m2.720s
```

This timing gets slower and slower with larger data sets:
```
time python expect.10mb.py
real	0m24.363s
user	0m8.750s
sys	0m14.540s
```

```
time python expect.25mb.py
real	2m13.992s
user	0m46.930s
sys	1m21.810s
```

Anything beyond 50MB is almost intolerably slow - taking 30 minutes+ to finish on my cir. 2015 Macbook Pro (while finishing in < 30 seconds when run by hand in a terminal emulator).

`pexpect` has a configurable `maxread` that can be used to increase the number of bytes read at a time:
https://github.com/pexpect/pexpect/blob/master/pexpect/pty_spawn.py#L34

...but the pty seems to impose an OS-dependent upper limit (looks like 2k on my Macbook Pro), so simply bumping `maxread` doesn't help.

ryanpetrello:
These two issues seem to be very closely related, and are possibly duplicates:

https://github.com/pexpect/pexpect/issues/385
https://github.com/pexpect/pexpect/issues/352

ryanpetrello:
I passed this through cProfile and found that almost all of the time spent seems to be in the new data consumption/search:

[...]

Issue #385, opened 2016-10-19
pexpect with long stdout - maxread and searchwindowsize example


aavella77:
Hi,

Environment:
pexpect (2.3)
python 2.7.5
centos 7

I am trying a pexpect script that throws large amounts of data to stdout.  I see that the display on the screen is very slow and takes more than 3 to 4 hours to come back with the prompt.  I have read the documentation but I cannot figure out how to configure the maxread and searhwindowsize parameters.  Are there examples on the Internet? or can someone provide them?  Is this a bug?

```
child = pexpect.spawn('ssh %s@%s' % (admin_username, mgmnt_ip), env = os.environ, maxread=100000000, searchwindowsize=2000)

cmd = 'python my-script.py -c config.json'  
if not cli.run_in_session(child,cmd,prompt,28800):
    assert 0, "%s command failed" % cmd 
```

Thanks,
Alejandro

ohmkay:
I have the same problem.  As far as I can tell, your maxread parameter is setup correctly.  The default is 2000 characters.  I've used a large number like yourself, but it appears the maximum size of this parameter is 4092.  Not nearly enough for my usage and very disappointing.  I've read somewhere it may have something to do with the TTY settings but I have found no work arounds.

MengJueM:
I also faced the same issue and am watching the good solutions.

ryanpetrello:
@aavella77 @MengJueM I've written a PR that I think should drastically improve performance - mind giving it a shot?

https://github.com/takluyver

takluyver:
Closing as #464 was merged. Let me know if performance is still problematic in this case with Pexpect >= 4.4 - we can reopen the issue if necessary.

Issue #352, opened 2016-06-07
prompt method on pxssh.pxssh takes too long on "large output" (~14MByte)


AlonElharar1:
While running the following lines:
        connection.sendline("cat  file_size_of_14_megabytes.log")
        connection.prompt():

The connection.prompt() takes "too long":
"too long" is means that cat 14 megabytes takes about 150 seconds - ~10 seconds for 1 single Mega.
The "prompt" method eventually goes to "def expect_loop(self, timeout=-1):" method on "Expecter" class.
On the following lines from implementation there is several problems:

```
While True:
       idx = self.new_data(incoming)
       ...
       incoming = spawn.read_nonblocking(spawn.maxread, timeout)
       time.sleep(0.0001)
```
1. The spawn.maxread defined to 2048, is too small number, but that number as no affect since the limit on OSX-El Capitan is 1024. the spawn.maxread number has no affect even if it defined to millions.
2. On the while loop for each incoming(small size) we check for new data on "idx = self.new_data(incoming)" - and that takes a lot of time.

The solution:
The solution is to read nonblocking data with greater size, so the "new_data" method handle a bigger data.

The solution is on "expect_loop(self, timeout=-1)" method:

```
while True:
            ....
            # No match at this point
            if (timeout is not None) and (timeout < 0):
                    return self.timeout()
            # Still have time left, so read more data
       (-) incoming = spawn.read_nonblocking(spawn.maxread, timeout)
       (+) incoming = ""
       (+) for _ in range(30):
       (+)     try:
       (+)         incoming += spawn.read_nonblocking(spawn.maxread, 0.5)
       (+)     except Exception:
       (+)         break
           time.sleep(0.0001)
       ....
```
- The "magic numbers" (0.5, 30) should be constant/parameter.
- The expect Exception can replaced to Timeout exception.
- My mail is elharrar.alon@gmail.com

AlonElharar1:
Shrarpen the spoon will help me a lot. right now it's affected on smaller files too(size of ~3MB), slow user experience compared to terminal/shell direct connection, so sharpen the spoon will be great.

Thanks!

ryanpetrello:
@AlonElharar1 I've opened a PR that I think should drastically improve performance - mind giving it a shot?

https://github.com/takluyver

takluyver:
Closing as #464 was merged. Let me know if performance is still problematic in this case with Pexpect >= 4.4 - we can reopen the issue if necessary.

PR #464, opened 2018-01-15
optimize pty buffering and searching

ryanpetrello:
Python strings are slow and expensive as buffers because they're immutable;
replace the output buffer with a `StringIO/BytesIO` object

see: https://github.com/pexpect/pexpect/issues/438
see: https://github.com/pexpect/pexpect/issues/385
see: https://github.com/pexpect/pexpect/issues/352

### DIFF
--- a/pexpect/expect.py
+++ b/pexpect/expect.py
@@ -9,30 +9,42 @@ class Expecter(object):
         if searchwindowsize == -1:
             searchwindowsize = spawn.searchwindowsize
         self.searchwindowsize = searchwindowsize
-    
+
     def new_data(self, data):
         spawn = self.spawn
         searcher = self.searcher
 
-        incoming = spawn.buffer + data
-        freshlen = len(data)
-        index = searcher.search(incoming, freshlen, self.searchwindowsize)
+        pos = spawn._buffer.tell()
+        spawn._buffer.write(data)
+
+        # determine which chunk of data to search; if a windowsize is
+        # specified, this is the *new* data + the preceding <windowsize> bytes
+        if self.searchwindowsize:
+            spawn._buffer.seek(max(0, pos - self.searchwindowsize))
+            window = spawn._buffer.read(self.searchwindowsize + len(data))
+        else:
+            # otherwise, search the whole buffer (really slow for large datasets)
+            window = spawn.buffer
+        index = searcher.search(window, len(data))
         if index >= 0:
-            spawn.buffer = incoming[searcher.end:]
-            spawn.before = incoming[: searcher.start]
-            spawn.after = incoming[searcher.start: searcher.end]
+            value = spawn.buffer
+            spawn._buffer = spawn.buffer_type()
+            spawn._buffer.write(value[searcher.end:])
+            spawn.before = value[: searcher.start]
+            spawn.after = value[searcher.start: searcher.end]
             spawn.match = searcher.match
             spawn.match_index = index
             # Found a match
             return index
-    
-        spawn.buffer = incoming
-    
+        elif self.searchwindowsize:
+            spawn._buffer = spawn.buffer_type()
+            spawn._buffer.write(window)
+
     def eof(self, err=None):
         spawn = self.spawn
 
         spawn.before = spawn.buffer
-        spawn.buffer = spawn.string_type()
+        spawn._buffer = spawn.buffer_type()
         spawn.after = EOF
         index = self.searcher.eof_index
         if index >= 0:
@@ -83,7 +95,7 @@ class Expecter(object):
 
         try:
             incoming = spawn.buffer
-            spawn.buffer = spawn.string_type()  # Treat buffer as new data
+            spawn._buffer = spawn.buffer_type()
             while True:
                 idx = self.new_data(incoming)
                 # Keep reading until exception or return.
--- a/pexpect/spawnbase.py
+++ b/pexpect/spawnbase.py
@@ -1,3 +1,4 @@
+from io import StringIO, BytesIO
 import codecs
 import os
 import sys
@@ -57,8 +58,6 @@ class SpawnBase(object):
         self.logfile_send = None
         # max bytes to read at one time into buffer
         self.maxread = maxread
-        # This is the read buffer. See maxread.
-        self.buffer = bytes() if (encoding is None) else text_type()
         # Data before searchwindowsize point is preserved, but not searched.
         self.searchwindowsize = searchwindowsize
         # Delay used before sending data to child. Time in seconds.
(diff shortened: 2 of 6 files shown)


ryanpetrello:
Looks like I've got some py3 issues to work out, will address the failing tests.

ryanpetrello:
Example test case that illustrates the performance improvement:

```python
    def test_large_stdout_stream(self):
        e = pexpect.spawn('openssl rand -base64 {}'.format(1024*1024*25), searchwindowsize=1000, timeout=None)
        resp = e.expect(['Password:', pexpect.EOF, pexpect.TIMEOUT])
```

Before:
```shell
$ time py.test tests/test_performance.py -k large_stdout_stream
140.38s user 178.81s system 96% cpu 5:29.90 total
```

After:
```shell
$ time py.test tests/test_performance.py -k large_stdout_stream
2.74s user 0.89s system 51% cpu 6.984 total
```

ryanpetrello:
@takluyver as best I can tell, you're the current maintainer of pexpect?

takluyver:
Thanks. I think this makes sense, but I also think that there are almost certainly people relying on `spawn.buffer` being a bytes/string object, rather than a file-like object. I've already learned the hard way that with code as old as pexpect, it's hard to make any change without breaking things.

Is it practical to make a new private attribute like `spawn._buffer_io`, and then make buffer a property which returns `spawn._buffer_io.getvalue()`? That way the code in pexpect can do things efficiently, but code that gets the buffer as a string should still work.

(I'd probably also implement setting `spawn.buffer = '...'` - hopefully there's less code that does that, but it's not much extra work.)

ryanpetrello:
@takluyver Good point re: backwards compatibility - I've uploaded another commit that I think should address this.

pexpect/expect.py
```diff
@@ -9,30 +9,41 @@ def __init__(self, spawn, searcher, searchwindowsize=-1):
...
+            spawn.after = value[searcher.start: searcher.end]
             spawn.match = searcher.match
             spawn.match_index = index
             # Found a match
             return index
-    
-        spawn.buffer = incoming
-    
+        elif self.searchwindowsize:
+            spawn._buffer = spawn.buffer_type(window)
```
takluyver: I think this could do the wrong thing in some cases. I was just testing the BytesIO/StringIO constructors because I was going to suggest getting rid of the `.write()` on line 32, and I noticed that if you pass data into the constructor, it leaves the cursor at the beginning, so `.tell()` returns 0, and writing overwrites the data already there.

```python
import io
s = io.StringIO('abcd')
s.tell()  # -> 0
s.write('zz')
s.getvalue()  # -> 'zzcd'
```

pexpect/expect.py
```diff
@@ -263,25 +251,16 @@ def __str__(self):
         ss = list(zip(*ss))[1]
         return '\n'.join(ss)
 
-    def search(self, buffer, freshlen, searchwindowsize=None):
```
takluyver: I'm not sure that we can get rid of these parameters so easily - backwards compatibility again.

pexpect/spawnbase.py
```diff
@@ -140,6 +147,15 @@ def _coerce_send_string(self, s):
             return s.encode('utf-8')
         return s
 
+    def get_buffer(self):
```
takluyver: Let's make these private methods (`_` prefix), so people don't start using them directly and cause us another backwards compatibility headache.

pexpect/spawnbase.py
```diff
@@ -1,3 +1,8 @@
+try:
+    from StringIO import StringIO
```
takluyver: I think we should use `io.StringIO` in all cases, because that's strict about unicode, whereas `StringIO.StringIO` isn't.

takluyver: (By 'in all cases', I mean on Python 2 as well as Python 3 - of course we still need to use `io.BytesIO` when we're handling bytes)

pexpect/expect.py
ryanpetrello: Which constructor?  I'm not sure I follow this comment.

pexpect/expect.py
```diff
@@ -157,8 +168,8 @@ def __str__(self):
...
-        '''This searches 'buffer' for the first occurrence of one of the search
+    def search(self, buff, freshlen, searchwindowsize=None):
```
ryanpetrello: @takluyver related to your comment, I just reverted these - the changes I made ended up being unnecessary.

pexpect/spawnbase.py
```diff
@@ -140,6 +143,15 @@ def _coerce_send_string(self, s):
             return s.encode('utf-8')
         return s
 
+    def _get_buffer(self):
```
ryanpetrello: @takluyver private methods now

pexpect/spawnbase.py
```diff
@@ -1,3 +1,4 @@
+from io import StringIO, BytesIO
```
ryanpetrello: @takluyver import fixed

pexpect/expect.py
ryanpetrello: Oh, are you suggesting that I should instead do:

```python
spawn._buffer = spawn.buffer_type()
spawn._buffer.write(window)
```

?

pexpect/expect.py
takluyver: Great, thanks. Could you also revert `buff` back to `buffer` in the parameters? Any parameter in Python can be passed as a keyword argument, so people might be relying on the name.

pexpect/expect.py
takluyver: Yup, that's what I meant. That way, writing more data with `_buffer.write()` afterwards will add it to the end of the buffer, rather than overwriting existing data.

pexpect/expect.py
ryanpetrello: Good point, I'll fix that.

pexpect/spawnbase.py
```diff
@@ -140,6 +143,15 @@ def _coerce_send_string(self, s):
...
         return s
 
+    def _get_buffer(self):
+        return self._buffer.getvalue()
+
+    def _set_buffer(self, value):
+        self._buffer = self.buffer_type()
+        self._buffer.write(value)
+
+    buffer = property(_get_buffer, _set_buffer)
```
takluyver: Could we have a comment to remind anyone working on the code that this interface as a string is required for backwards compatibility?

ryanpetrello:
Sounds great, @takluyver.  Thanks for taking the time to review this!

ryanpetrello:
Hey @takluyver,

Any chance this is ready to merge?
Thanks!

takluyver:
Thanks!

PR #467, opened 2018-02-12
FIX: bad representation of bytes patterns.

JulienPalard:
I was getting:

    0: re.compile("b'^[^\r\n]*\r?\n'")
    1: re.compile("b'^--More-- or \\(q\\)uit'")
    2: re.compile("b'^\\(.*\\) #'")

So I decided to drop the `%s` and keep only the `%r`, which works equally for
str and bytes:

    >>> '    %d: re.compile(%r)' % (1, b'foo')
    "    1: re.compile(b'foo')"
    >>> '    %d: re.compile(%r)' % (1, 'foo')
    "    1: re.compile('foo')"

### DIFF
--- a/pexpect/expect.py
+++ b/pexpect/expect.py
@@ -158,7 +158,7 @@ class searcher_string(object):
         '''This returns a human-readable string that represents the state of
         the object.'''
 
-        ss = [(ns[0], '    %d: "%s"' % ns) for ns in self._strings]
+        ss = [(ns[0], '    %d: %r' % ns) for ns in self._strings]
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks, I agree that `%r` makes sense here. It looks like there are a couple of tests that need updating as a result - could you have  a look?

JulienPalard:
@takluyver Woops, test updated. I had to modify the repr in the string matcher for consistency.

takluyver:
Thanks, looks good to me.

PR #468, opened 2018-02-23
Add custom regexes and allow for delicate SSH servers

Red-M:
Patch that allows setting some custom regexes for pxssh and allows pxssh to be more gentle with target SSH servers and the shells they spawn.

This is for #451

Please let me know if I need to change anything in regards to docs or comments and those requests will be handled.

### DIFF
--- a/pexpect/pxssh.py
+++ b/pexpect/pxssh.py
@@ -274,7 +274,8 @@ class pxssh (spawn):
         '''
         
         session_regex_array = ["(?i)are you sure you want to continue connecting", original_prompt, self.password_regex, "(?i)permission denied", "(?i)terminal type", TIMEOUT]
-        session_init_regex_array = session_regex_array
+        session_init_regex_array = []
+        session_init_regex_array.extend(session_regex_array)
         session_init_regex_array.extend(["(?i)connection closed by remote host", EOF])
 
         ssh_options = ''.join([" -o '%s=%s'" % (o, v) for (o, v) in self.options.items()])
--- a/pexpect/pxssh.py
+++ b/pexpect/pxssh.py
@@ -274,8 +274,8 @@ class pxssh (spawn):
         '''
         
         session_regex_array = ["(?i)are you sure you want to continue connecting", original_prompt, self.password_regex, "(?i)permission denied", "(?i)terminal type", TIMEOUT]
-        session_init_regex_array = session_regex_array.copy()
-        session_init_regex_array.append("(?i)connection closed by remote host", EOF)
+        session_init_regex_array = session_regex_array
+        session_init_regex_array.extend(["(?i)connection closed by remote host", EOF])
 
         ssh_options = ''.join([" -o '%s=%s'" % (o, v) for (o, v) in self.options.items()])
         if quiet:
--- a/pexpect/pxssh.py
+++ b/pexpect/pxssh.py
@@ -273,9 +273,9 @@ class pxssh (spawn):
         manually set the :attr:`PROMPT` attribute.
         '''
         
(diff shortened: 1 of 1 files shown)


Red-M:
Is there an issue with this PR @takluyver ?

I have to monkey patch this to fix devices that include password in the MOTD after a successful login.

takluyver:
Sorry, I just hadn't looked at it. I've never used pxssh myself, so I tend to drag my feet on PRs for it, just because it doesn't interest me much.

Does PR #452 help to avoid issues with "password" appearing in the MOTD? It sounds like @orulz was running into the same thing.

pexpect/pxssh.py
```diff
@@ -95,7 +95,9 @@ class pxssh (spawn):
 
     def __init__ (self, timeout=30, maxread=2000, searchwindowsize=None,
                     logfile=None, cwd=None, env=None, ignore_sighup=True, echo=True,
-                    options={}, encoding=None, codec_errors='strict'):
+                    options={}, encoding=None, codec_errors='strict',
+                       password_regex='(?i)(?:password:)|(?:passphrase for key)',
+                       disable_original_sync=False):
```
takluyver: Style point: I prefer to avoid negative boolean parameters like `disable_foo`, unless there's a clear reason that the negative makes sense.

Looking at the rest of this class, I think the parameter fits better in the `login()` method (which already has parameters affecting syncing and changing the prompt). I'd probably just call it `sync_original_prompt`.

pexpect/pxssh.py
```diff
@@ -195,7 +208,10 @@ def sync_original_prompt (self, sync_multiplier=1.0):
         can take 12 seconds. Low latency connections are more likely to fail
         with a low sync_multiplier. Best case sync time gets worse with a
         high sync multiplier (500 ms with default). '''
-
+        
+        if self.disable_original_sync:
+            return(True)
```
takluyver: Also, rather than making the method a no-op, let's check it in `login()` and skip calling the method.

Red-M:
This all seems fine to me.

Red-M:
While PR #452 does fix it, for some reason this was reverted due to other devices not matching the password prompt (iirc, brocade devices).

I made this PR because devices like PFsense have `password` in the custom shell script that the devices spawns on login (it is also the reason to disable `sync_original_prompt` because sending a newline also counts as a `0` which is to exit/logout).

So I figured I might as well write something to fix both.

Red-M:
> Sorry, I just hadn't looked at it. I've never used pxssh myself, so I tend to drag my feet on PRs for it, just because it doesn't interest me much.

Do you need someone to look after pxssh a bit more?

takluyver:
Thanks, this looks good now. :-)

Having a maintainer who's more familiar with pxssh would certainly be good. If you're interested, would you mind sending us (the current maintenance team is myself, @jquast and @MountainRider ) a bit of information about yourself, so we know who we're trusting.

Red-M:
Let me know where and what to send. :)

takluyver:
You can post here if you're happy to do it in public, or use the email addresses on our Github profiles.

As for what, there's no fixed list, I'd just like to have an idea who you are. What's your name? Where do you live (roughly - I don't need an address)? How do you come to be using pxssh a lot? What else are you interested in or working on?

Red-M:
I'll send an email around.

Thanks!

takluyver:
@MountainRider my email to your technicolor.com address bounced. Do you have another address we should use now?

takluyver:
There were no objections, so I've added @Red-M on GitHub. Welcome to Pexpect maintenance! :cookie:

Red-M:
Thanks!

PR #474, opened 2018-03-23
Allow callers to enable to use of select.poll()

cooperlees:
- Add the use_poll bool for callers to choose if they want poll
- This ensure we leave the default as it is today
- Added unit tests to use select.poll()
- Updated documentation guessing format etc.

If wrong I can fix, or if you could just fix post merge that would be amazing :D

### DIFF
--- a/doc/history.rst
+++ b/doc/history.rst
@@ -7,10 +7,10 @@ Releases
 Version 4.5
 ```````````
 
-* :class:`~.spawn` now has a ``use_poll`` parameter. This allows the use
-  of select.poll() on fds
-* :class:`~.fdspawn` now has a ``use_poll`` paramneter. This allow the use
-  of select.poll() on fds
+* :class:`~.spawn` + `~.fdspawn` now have an ``use_poll`` parameter.
+  This change allows the use of select.poll() on file descriptors.
+  It allows for > 1024 file descriptors on the system, and, it is not used by
+  default due to compatibility concerns and must be explicitly enabled.
 
 Version 4.4
 ```````````
--- a/doc/history.rst
+++ b/doc/history.rst
@@ -4,6 +4,14 @@ History
 Releases
 --------
 
+Version 4.5
+```````````
+
+* :class:`~.spawn` now has a ``use_poll`` parameter. This allows the use
+  of select.poll() on fds
+* :class:`~.fdspawn` now has a ``use_poll`` paramneter. This allow the use
+  of select.poll() on fds
+
 Version 4.4
 ```````````
 
@@ -295,4 +303,3 @@ Moves and forks
   to call the new fork Pexpected. Noah Spurrier agreed to let them use
   the name Pexpect, so Pexpect versions 3 and above are based on this
   fork, which now lives `here on Github <https://github.com/pexpect/pexpect>`_.
-
--- a/pexpect/utils.py
+++ b/pexpect/utils.py
@@ -154,3 +154,32 @@ def select_ignore_interrupts(iwtd, owtd, ewtd, timeout=None):
                 # something else caused the select.error, so
                 # this actually is an exception.
                 raise
+
+
+def poll_ignore_interrupts(fds, timeout=None):
+    """Simple wrapper around poll for a list of file descriptors."""
+
+    if timeout is not None:
+        end_time = time.time() + timeout
+
+    poller = select.poll()
+    for fd in fds:
+        poller.register(fd)
+        while True:
+            try:
+                timeout_ms = None if timeout is None else timeout * 1000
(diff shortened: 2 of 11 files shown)


Red-M:
Can I ask that if you're updating the release changelog that I can put pxssh's release changes in first?
I have a few new features planned and I don't want to break your PR here.

cooperlees:
I don't 100% follow what you want me to do? Just wait and rebase yes? I guess I should also get the test coverage back up.

pexpect/utils.py
```diff
@@ -154,3 +154,32 @@ def select_ignore_interrupts(iwtd, owtd, ewtd, timeout=None):
...
+def poll_ignore_interrupts(fds, timeout=None):
+    """Simple wrapper around poll for a list of file descriptors."""
+
+    if timeout is not None:
+        end_time = time.time() + timeout
+
+    poller = select.poll()
+    for fd in fds:
+        poller.register(fd)
+        while True:
```
takluyver: This should probably be outside the for loop, if I'm understanding it correctly.

takluyver:
This looks OK to me.

Release notes are always a pain when there are multiple changes in flight at the same time, because they guarantee conflicts. Tools like [towncrier](https://pypi.python.org/pypi/towncrier) can help, but it's probably overkill to use that for a relatively quiet project like Pexpect.

Red-M:
I'm ok to merge this.
I'll do another PR to add my release notes in for pxssh :)

pexpect/utils.py
Red-M: Before I merge this, can you confirm that this is what you want to do?

cooperlees: I want to register each fd right or am I misunderstanding what I need to do here? I guess we're only really ever passing 1 FD tho ...

Red-M: you never break the while loop and it has no condition to allow the next `fd` to come into play.

takluyver: I think the `while` loop should be dedented one level so it happens after the `for` loop instead of inside it.

It doesn't make a difference when we only pass one fd, but it will be broken if the function is ever used with more than one, so we might as well get it right.

cooperlees:
changed the API to take one fd only more like select_ignore_interrupts

Red-M:
Seems fine to me, if @takluyver has no comments I'll merge this if he doesn't.

doc/history.rst
```diff
@@ -4,6 +4,14 @@ History
 Releases
 --------
 
+Version 4.5
+```````````
+
+* :class:`~.spawn` now has a ``use_poll`` parameter. This allows the use
+  of select.poll() on fds
+* :class:`~.fdspawn` now has a ``use_poll`` paramneter. This allow the use
+  of select.poll() on fds
```
takluyver: Let's combine these two notes into one (it's one conceptual change, even if it affects two classes) and mention that:

- It lets you get round problems with file descriptors > 1024.
- It's not used by default because of compatibility concerns, so the downstream code has to decide to use it.

pexpect/utils.py
```diff
@@ -154,3 +154,31 @@ def select_ignore_interrupts(iwtd, owtd, ewtd, timeout=None):
                 # something else caused the select.error, so
                 # this actually is an exception.
                 raise
+
+
+def poll_ignore_interrupts(fd, timeout=None):
+    """Simple wrapper around poll for a list of file descriptors."""
```
takluyver: Don't forget to update the docstring now that it only takes one fd.

doc/history.rst
takluyver: Ping @cooperlees ? If you can make these changes to the docs, then I think the PR is ready to merge.

cooperlees:
Updated documentation. I do feel it would of been faster for you to just merge and fix up the documentation to how you want it. A new comer is never going to get documentation how main mergers like it. I hope it's all fine now.

takluyver:
@cooperlees sorry,  I just noticed that there's a problem in this code - in `__interact_copy` it needs to handle two fds (like the select implementation does). Have you got time to take a look?

cooperlees:
Will take a look tomorrow.

MountainRider:
It's important for contributors to make the necessary documentation changes. When someone has a question about the documentation, they will look at the `git blame` output and contact the person who committed the changes. Also, while it might be faster for someone else to make the changes, contributors won't learn what is expected if they don't make the changes themselves.

cooperlees:
https://github.com/pexpect/pexpect/pull/485

ryanpetrello:
It doesn't seem like the implementation respects the `timeout` argument to `spawn()`: https://github.com/pexpect/pexpect/issues/491

Issue #478, opened 2018-03-27
'spawn.before' is truncated when using a 'searchwindowsize'


alirez:
When `searchwindowsize` is non-default, `spawn.before` no longer contains all the text up to the matched pattern.

The behavior seems to have changed after fd7332f5. Below is a quick demonstration of the issue.

The spawned shell prints out 500 lines and then the word PATTERN. It worked as expected before the above commit, i.e. 500 lines where in `.before` after the pattern matched.
```
$ git checkout fd7332f5~1 >& /dev/null
$ python
Python 2.7.13 (default, Nov 24 2017, 17:33:09)
[GCC 6.3.0 20170516] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import pexpect
>>> child = pexpect.spawn('''/bin/bash -c "openssl rand -base64 {} | head -500 | nl --number-format=rz --number-width=5 2>&1 ; echo 'PATTERN'"'''.format(1024 * 1024 * 2), searchwindowsize=128)
>>> child.expect(['PATTERN'])
0
>>> print(len(child.before.splitlines()))
500
```

However, on fd7332f5:
```
$ git checkout fd7332f5 >& /dev/null
$ python
Python 2.7.13 (default, Nov 24 2017, 17:33:09)
[GCC 6.3.0 20170516] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import pexpect
>>> child = pexpect.spawn('''/bin/bash -c "openssl rand -base64 {} | head -500 | nl --number-format=rz --number-width=5 2>&1 ; echo 'PATTERN'"'''.format(1024 * 1024 * 2), searchwindowsize=128)
>>> child.expect(['PATTERN'])
0
>>> print(len(child.before.splitlines()))
8
```
Only a fraction of the actual output is stored in `.before`

takluyver:
@ryanpetrello could you take a look? It seems like this might be related to your PR #464. Thanks!

ryanpetrello:
@takluyver this does look related to my change; the issue is that the new implementation throws away the anything beyond the buffer + search window on each chunk of data that's read, and as a result, `spawn.before` doesn't have the context it needs.  I'm going to work on a PR that uses an extra string/bytes buffer to track `before` content _up to_ a match.

ryanpetrello:
@takluyver I think there's actually _another_ bug hidden here in the `after` attribute.  I'm going to write some tests and experiment.

takluyver:
[...]

PR #479, opened 2018-03-27
fix a bug causing 'spawn.before' to be truncated w/ searchwindowsize use

ryanpetrello:
see: https://github.com/pexpect/pexpect/issues/478

### DIFF
--- a/pexpect/expect.py
+++ b/pexpect/expect.py
@@ -16,6 +16,7 @@ class Expecter(object):
 
         pos = spawn._buffer.tell()
         spawn._buffer.write(data)
+        spawn._before.write(data)
 
         # determine which chunk of data to search; if a windowsize is
         # specified, this is the *new* data + the preceding <windowsize> bytes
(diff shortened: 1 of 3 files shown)


pexpect/expect.py
```diff
@@ -27,11 +28,11 @@ def new_data(self, data):
             window = spawn.buffer
         index = searcher.search(window, len(data))
         if index >= 0:
-            value = spawn.buffer
             spawn._buffer = spawn.buffer_type()
-            spawn._buffer.write(value[searcher.end:])
-            spawn.before = value[: searcher.start]
-            spawn.after = value[searcher.start: searcher.end]
+            spawn._buffer.write(window[searcher.end:])
```
ryanpetrello: @takluyver these search match indexes are _actually_ relative to `window`, which _may not_ be equal to `spawn.buffer` when `searchwindowsize` is in use.

takluyver:
Thanks for the investigation and speedy fix. I think what you're doing here makes sense.

I don't quite know what you mean about other bugs - do you think they're fixed by this PR? Is there something more you still want to do? I'm happy to merge this if you're happy with it.

ryanpetrello:
> I don't quite know what you mean about other bugs

This PR _also_ changes how I set `spawn.after` to use `window` instead of `value`; I think without this, people will _also_ see bugs on that property similar to what https://github.com/pexpect/pexpect/issues/478 describes for `before`.

takluyver:
Aha, gotcha. Merging now, thanks!

PR #485, opened 2018-04-13
Make poll_ignore_interrupts register multiple FDs

cooperlees:
Like select.select() call and fix broken __interact_copy implementation


takluyver:
Thanks @cooperlees, that's precisely what I meant :-)

Issue #489, opened 2018-05-12
How to use .login with a ssh key?


kw-andy:
I'm trying to connect to a ssh server on which I have set a public key.

I have my private keys on the folder ` /home/XXX/priv_keys` but I'm not sure on how to use it with the `login` module.

The doc gives me this 

```
Set ssh_key to True to force passing the current SSH authentication socket to the to the desired hostname.
```

But eventually, pexpect still ask me for the password, which is deactivated on my distant server.

This is the code I've tried

```
from pexpect import pxssh
import getpass
try:
    s = pxssh.pxssh()
    hostname = '1xx.138.12.xx'
    username = 'and_wk'
    s.login(hostname, username, '/home/XXX/priv_keys', True) # the path with the private keys has been used instead of the password and the ssh_key has been set to True  
    s.sendline('uptime')   # run a command
    s.prompt()             # match the prompt
    print(s.before.decode())        # print everything before the prompt.](url)
```

Thanks in advance for any help

Red-M:
That option is for the socket not a key.
To use the option load a key into a socket based ssh key agent and either:

- Run the script from a bash line with said socket agent active
- Pass the socket agent to pxssh via the ssh options.

An example would be:
On your bash command line:
```
ssh-add <ssh key path>
./pxssh_script.py
```
or loading the agent from another session under the same user within pxssh:
bash command line:
```
ssh-add <ssh key path>
echo $SSH_AUTH_SOCK
```
Echo'd line is the path to the socket of that instance of the ssh key agent.
Python:
```python
from pexpect import pxssh
import getpass
try:
    ssh_options = {'IdentityAgent': '<agent path>'}
    s = pxssh.pxssh(options=ssh_options)
    hostname = '1xx.138.12.xx'
    username = 'and_wk'
    s.login(hostname, username)
    s.sendline('uptime')   # run a command
    s.prompt()             # match the prompt
    print(s.before.decode())        # print everything before the prompt.](url)
```


With all the said above you wouldn't need to load in your SSH key so instead:
```python
s.login(hostname, username)
```

Red-M:
To answer with something that could help you rather than explain docs.
Can you confirm that you can login to the remote machine with said key and that the remote side has ~/.ssh set to the correct permissions?

Another common issue is that the SSH server does not have keys set up correctly in the terms that the server is looking for the keys in the wrong place or its not configured to look for keys.

kw-andy:
Hi Red,

OMG.

[...]

Issue #429, opened 2017-07-10
Feature Request: Pass config_file to pxssh instead of user


cognifloyd:
For now, I'm using `pexpect.spawn` to handle some fairly standard ssh connections. I'd like to use `pexpect.pxssh` to benefit from handling variations in prompts, passing in the password, etc. But, `pxssh` requires a user, and it doesn't let me specify an ssh config file.

I have an auto-generated ssh config file that has all of the connection info I want to use (for a lot of hosts), including user.
Would you extend `pxssh` so that I can set a config file for the connection (the `-F` option of `ssh`)? When you do that, can you make the user optional, and maybe raise an exception if neither config file nor user are specified?

jnatschev:
Hi.

I just did a test and wanted to ask whether the ‘-F’ bit your asking for is needed?

The reason i ask is because if your config file is named ‘config’ and is located in your '$HOME/.ssh’ directory, then the ‘-F’ bit isn’t needed.

To the ‘pexpect’ people, I haven’t really learned how to use GitHub properly yet. The changes I made to the pxssh.py file are:
Original:
def login (self, server, username, password='', terminal_type='ansi',
            original_prompt=r"[#$]", login_timeout=10, port=None,
            auto_prompt_reset=True, ssh_key=None, quiet=True,
            sync_multiplier=1, check_local_ip=True):

Modified:
def login (self, server, username=None, password='', terminal_type='ansi',
            original_prompt=r"[#$]", login_timeout=10, port=None,
            auto_prompt_reset=True, ssh_key=None, quiet=True,
            sync_multiplier=1, check_local_ip=True):

Original:
cmd = "ssh %s -l %s %s" % (ssh_options, username, server)

Modified:
if not username:
    cmd = 'ssh {} {}'.format(ssh_options, server)
else:
    cmd = "ssh %s -l %s %s" % (ssh_options, username, server)

Not sure if this is satisfactory enough, though it does work using the “standard” $HOME/.ssh/config file without needing to provide an ssh “-F” argument.

PS. Sorry I am not confident with the GitHub side of stuff yet.


Cheers,
John.

cognifloyd:
I have an auto generated ssh config file full of servers that users can refer to system wide. However, that file is not part of there ~/.ssh/config file by default, and ssh is too old on one server (ugh) to support the Include directive. So, yes `-F` is very important.

[...]

PR #490, opened 2018-05-20
Update docs and add SSH config file passing

Red-M:
As per the title, this is for issues #489 and #429.

@takluyver it might be a good idea to do a docs update from commit 6b741f9 as this will solve a small piece of miscommunication I introduced by doing doc updates for my own additions but not for pre-existing features.

### DIFF
--- a/pexpect/pxssh.py
+++ b/pexpect/pxssh.py
@@ -299,7 +299,7 @@ class pxssh (spawn):
         Set ``ssh_key`` to `True` to force passing the current SSH authentication socket
         to the desired ``hostname``.
         
-        Setting ``ssh_config`` to a file path of an SSH client config file will pass that
+        Setting ``ssh_config`` to a file path string of an SSH client config file will pass that
         file to the client to handle itself. You may set any options you wish in here, however
         doing so will require you to post extra information that you may not want to if you
         run into issues.
--- a/pexpect/pxssh.py
+++ b/pexpect/pxssh.py
@@ -259,7 +259,7 @@ class pxssh (spawn):
                 sync_multiplier=1, check_local_ip=True,
                 password_regex=r'(?i)(?:password:)|(?:passphrase for key)',
                 ssh_tunnels={}, spawn_local_ssh=True,
-                sync_original_prompt=True):
+                sync_original_prompt=True, ssh_config=None):
         '''This logs the user into the given server.
 
         It uses
@@ -312,6 +312,14 @@ class pxssh (spawn):
             ssh_options = ssh_options + " -o'NoHostAuthenticationForLocalhost=yes'"
         if self.force_password:
             ssh_options = ssh_options + ' ' + self.SSH_OPTS
+        if ssh_config is not None:
+            if spawn_local_ssh:
+                try:
+                    if spawn_local_ssh:
+                        os.path.isfile(ssh_config)
+                except:
+                    raise ExceptionPxssh('SSH config does not exist')
+                ssh_options = ssh_options + ssh_config
         if port is not None:
             ssh_options = ssh_options + ' -p %s'%(str(port))
         if ssh_key is not None:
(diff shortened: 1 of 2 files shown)


pexpect/pxssh.py
```diff
@@ -294,8 +294,15 @@ def login (self, server, username, password='', terminal_type='ansi',
         session to do so. Setting this option to `False` and not having an active session
         will trigger an error.
         
-        Set ``ssh_key`` to `True` to force passing the current SSH authentication socket to the
+        Setting ``ssh_key`` to a file path to an SSH private key will use that SSH key
```
takluyver: I'd like the docstring to be consistent about using either "Set ... to ..." or "Setting ... will ...". I don't have a strong preference which (though it's easier to change the bits you've just added), but it should use the same one throughout.

pexpect/pxssh.py
```diff
@@ -310,6 +317,13 @@ def login (self, server, username, password='', terminal_type='ansi',
             ssh_options = ssh_options + " -o'NoHostAuthenticationForLocalhost=yes'"
         if self.force_password:
             ssh_options = ssh_options + ' ' + self.SSH_OPTS
+        if ssh_config is not None:
+            try:
+                if spawn_local_ssh:
+                    os.path.isfile(ssh_config)
+            except:
```
takluyver: Catching all exceptions like this is not good practice, because it can obscure other problems. In this case it also won't do what you want, because `isfile()` returns False rather that raising an exception.

Maybe you want:

```python
if spawn_local_ssh and not os.path.isfile(ssh_config):
    raise ...
```

tests/test_pxssh.py
```diff
@@ -85,6 +85,13 @@ def test_remote_ssh_tunnel_string(self):
         if confirmation_strings!=len(confirmation_array):
             assert False, 'String generated from remote tunneling is incorrect.'
 
+    def test_ssh_config_passing_string(self):
+        ssh = pxssh.pxssh(debug_command_string=True)
+        config_path = '/fakepath/fake/config_file'
```
takluyver: And once the file-exists check is working, you presumably need to create a real named temp file to test with.

It would also be good to have a test with a nonexistent file to check that it errors.

takluyver:
I'm not 100% sure what you mean about a docs update, but if you mean rebuilding them, there's good news! Readthedocs rebuilds docs automatically whenever changes are made, so it will update once this is merged.

pexpect/pxssh.py
Red-M: Sure thing.

pexpect/pxssh.py
Red-M: Odd because I copied what was already there.
I'll push this into a function as well as I debating either to leave it as it is or functionize it.

tests/test_pxssh.py
Red-M: yeap.

Red-M:
I'll fix these on the weekend and its good to hear about the docs at least!

Red-M:
Sorry for not getting back to this sooner however this should be up to spec now.

pexpect/pxssh.py
```diff
@@ -294,8 +294,15 @@ def login (self, server, username, password='', terminal_type='ansi',
         session to do so. Setting this option to `False` and not having an active session
         will trigger an error.
         
-        Set ``ssh_key`` to `True` to force passing the current SSH authentication socket to the
+        Set ``ssh_key`` to a file path to an SSH private key will use that SSH key
+        for the session authentication.
+        Set ``ssh_key`` to `True` to force passing the current SSH authentication socket
         to the desired ``hostname``.
+        
+        Set ``ssh_config`` to a file path string of an SSH client config file will pass that
```
takluyver: These need another minor tweak to fix the grammar for "Set" rather than "Setting":

- Setting X to Y **will** do Z
- Set X to Y **to** do Z

Red-M: Fair enough.

Red-M:
Fixed as per request.

PR #493, opened 2018-05-22
Make `crlf` match the default platform behaviour in `PopenSpawn`.

### LINKED ISSUE #488: Wrong value for crlf when using PopenSpawn
When using `PopenSpawn`, `crlf` should be set depending on the OS. However there are two issues in the current code:

- `crlf` is hardcoded to `\n` in the declaration of `PopenSpawn`, which is wrong on Windows.
- but anyway `crlf` is overwritten in the `SpawnBase` constructor so it is always set to `\r\n`, which is wrong on Unix systems.

jeberger:
Fixes #488

### DIFF
--- a/pexpect/popen_spawn.py
+++ b/pexpect/popen_spawn.py
@@ -18,17 +18,24 @@ from .exceptions import EOF
 from .utils import string_types
 
 class PopenSpawn(SpawnBase):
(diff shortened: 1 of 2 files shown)


takluyver:
Seems reasonable, thanks.

PR #501, opened 2018-06-12
replwrap: add async support

ThomasFaivre:
Add an 'async_' argument to run_command. When True, a Future object will
be returned which *must* be awaited in asynchronous code.

Signed-off-by: Robin Jarry <robin.jarry@6wind.com>
Signed-off-by: Thomas Faivre <thomas.faivre@6wind.com>

### DIFF
--- a/pexpect/replwrap.py
+++ b/pexpect/replwrap.py
@@ -61,11 +61,11 @@ class REPLWrapper(object):
         self.child.expect(orig_prompt)
         self.child.sendline(prompt_change)
 
-    def _expect_prompt(self, timeout=-1):
+    def _expect_prompt(self, timeout=-1, async_=False):
         return self.child.expect_exact([self.prompt, self.continuation_prompt],
-                                       timeout=timeout)
+                                       timeout=timeout, async_=async_)
 
-    def run_command(self, command, timeout=-1):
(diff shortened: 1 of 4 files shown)


pexpect/replwrap.py
```diff
@@ -61,11 +61,48 @@ def set_prompt(self, orig_prompt, prompt_change):
...
+        else:
+            result = asyncio.Future(loop=loop)
+        cmd_count = len(cmdlines)
+        out = []
+
+        # Called recursively through callbacks
+        def _run_line(fut=None):
+            if len(cmdlines) < cmd_count:
+                # gather output after the first line has been sent
+                out.append(self.child.before)
```
takluyver: This async recursive structure is not very easy to follow, in my opinion. Would it be better written as a coroutine?

ThomasFaivre:
Thanks for the very good point @takluyver 
We thought we needed that because we couldn't use "yield from" in replwrap.py but we can simply use the _async module...
That looks way nicer!

ThomasFaivre:
Hello!
Just a quick status update. How soon can I expect this in the master branch? Is there anything I need to add?

Thanks!

takluyver:
Looks good to me, thanks.

PR #502, opened 2018-06-21
Allow the ssh executable to be overridden

AlexThurston:
It's possible that you might want to use pxssh but with a custom ssh binary.  This adds the ability to overridden where the `ssh` binary is located.

### DIFF
--- a/pexpect/pxssh.py
+++ b/pexpect/pxssh.py
@@ -118,7 +118,7 @@ class pxssh (spawn):
     def __init__ (self, timeout=30, maxread=2000, searchwindowsize=None,
                     logfile=None, cwd=None, env=None, ignore_sighup=True, echo=True,
                     options={}, encoding=None, codec_errors='strict',
-                    debug_command_string=False):
+                    debug_command_string=False, overridden_ssh_cmd=None):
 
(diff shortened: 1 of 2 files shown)


Red-M:
I'd suggest that if you want to run a different binary that you modify the path for your system to include your binary rather than the default.

You can see an example of path overriding in the test suite, [here](https://github.com/pexpect/pexpect/blob/master/tests/test_pxssh.py#L9-L13).
If you want to be able to see where the normal ssh binary is for your platform you can use this code to do so.
```python
import subprocess
print(subprocess.check_output(['which','ssh']))
```

AlexThurston:
Yup. That’s certainly another option. I just thought that it might be nice to change the version from within the python process rather than for the entire environment and then start python.  The use case I was thinking of was making an application the used Pexpect and also packaged up the special ssh binary which would be referenced within the application.  This would allow the application to call into its own ssh binary without requiring the end user changing anything.

Red-M:
The test suite does change the binary to its own for the purposes of testing the library, from within python and when its done, the test suite changes it back to the original path.

So if people want something non-standard they can still have that with no code change on pxssh.

AlexThurston:
Fair enough.  This change isn't really all that necessary.

Issue #505, opened 2018-06-24
Support Windows better/fully using WinPTY


segevfiner:
It's possible to achieve better/full Windows support using [Winpty](https://github.com/rprichard/winpty) which emulates a pty on Windows using "trickery". 

This will allow making the full API of pexpect available on Windows without having to maintain the low level details of such a hack inside pexpect itself.

Whether the maintainers of pexepct will be welcome to merge such a thing into pexpect itself is a different question...

P.S. There are Python binding for WinPTY floating around such as [PyWinpty](https://github.com/spyder-ide/pywinpty)

takluyver:
I'm not interested in adding any new features to pexpect at this point - I've tried it before, and it's been a maintenance headache. I'm also very wary of a complex library that aims to hack a fundamental Unix feature onto Windows, though I understand that at least some people have had good results with it.

Personally, I'd like to see Pexpect superseded by something that separates the pattern-expecting machinery from the complexities of I/O, so it can automatically work with any source of data (ptys, winpty, pipes, sockets...), without needing explicit support for those. I'm trying to resist the urge to write that myself, because I have enough projects on the go already, but if anyone's willing to maintain it, I could prototype it and hand it over.

(For context, I'm one of ~4 active maintainers. I won't stop the other maintainers if they want to do this.)

Red-M:
I'd rather receive a PR for this as I am not 100% with pexpect's main code.

I'm interested in the separation of IO and pexpect though and I might take a crack at that sometime soon.

takluyver:
> I'm interested in the separation of IO and pexpect though and I might take a crack at that sometime soon.

To be clear, I'd suggest that this is a new project with a different name. That way there's no risk of breaking people's existing code, and no need to preserve Pexpect's rather dated API.

jquast:
I’ve stopped using pexpect... I use paramikko for ssh, and streamexpect for serial conversations.. much cleaner patterns and final code is more exact in error capturing.

All I think about contributing to pexpect anymore is documentation about what else to use :)

takluyver:
[streamexpect](https://github.com/digidotcom/python-streamexpect) did look promising for this for a while, but development seems to have stopped. Maybe the answer is for someone to ensure that's maintained. Or maybe it just works and doesn't need any more changes :grinning: .

[...]

Issue #414, opened 2017-04-23
pexpect.popen_spawn.PopenSpawn(ssh commend) didn't work on windows


whtis:
hello,
I am trying to use pexpect to Login SSH automatically on `windows10`.The python version is `3.5.2` and pexpect is `4.2.1`.
```
import traceback
import pexpect
from pexpect.popen_spawn import PopenSpawn

try:
    child = pexpect.popen_spawn.PopenSpawn('ssh root@192.168.21.225')
    a = child.expect(['password:', 'The authenticity of host'], timeout=300, async=True)
    if a == 0:
        child.sendline('123456789')
        print('this ip has exists in know_host files!')
    if a == 1:
        print('this ip will be added to know_host files!')
        child.sendline('yes')
        child.expect('password:')
        child.sendline('123456789')
        # send test
        child.sendline("df -h")
        child.interact()
except pexpect.EOF:
    traceback.print_exc()
```
because cmd can not use `ssh` I use the [`cmder`](https://github.com/cmderdev/cmder) instead. The `python` or `ssh` can work well if I typed artificial. But when I run the code in `cmder` like this:
`python ssh.py`

I got the error below and I can't find more information via google.

```
Traceback (most recent call last):
  File "C:\develop_tools\Python3\lib\site-packages\pexpect\expect.py", line 99, in expect_loop
    incoming = spawn.read_nonblocking(spawn.maxread, timeout)
  File "C:\develop_tools\Python3\lib\site-packages\pexpect\popen_spawn.py", line 66, in read_nonblocking
    raise EOF('End Of File (EOF).')
pexpect.exceptions.EOF: End Of File (EOF).

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "ssh.py", line 7, in <module>
    a = child.expect(['password:', 'The authenticity of host'])
  File "C:\develop_tools\Python3\lib\site-packages\pexpect\spawnbase.py", line 321, in expect
    timeout, searchwindowsize, async)
  File "C:\develop_tools\Python3\lib\site-packages\pexpect\spawnbase.py", line 345, in expect_list
    return exp.expect_loop(timeout)
  File "C:\develop_tools\Python3\lib\site-packages\pexpect\expect.py", line 105, in expect_loop
    return self.eof(e)
  File "C:\develop_tools\Python3\lib\site-packages\pexpect\expect.py", line 50, in eof
    raise EOF(msg)
pexpect.exceptions.EOF: End Of File (EOF).
[...]

Issue #339, opened 2016-03-15
pxssh import issue


rubycarlin:
import pxssh

Traceback (most recent call last):
  File "<pyshell#1>", line 1, in <module>
    import pxssh
ImportError: No module named pxssh

from pexpect import pxssh

Traceback (most recent call last):
  File "<pyshell#3>", line 1, in <module>
    from pexpect import pxssh
  File "C:\Python27\lib\site-packages\pexpect-4.0.1-py2.7.egg\pexpect\pxssh.py", line 23, in <module>
    from pexpect import ExceptionPexpect, TIMEOUT, EOF, spawn
ImportError: cannot import name spawn

takluyver:
pxssh is not currently supported on Windows, as it uses a pty.

rubycarlin:
Can you tell me what is the alternative to pxssh on windows !!

Thanks in advance.

takluyver:
[Paramiko](https://github.com/paramiko/paramiko), maybe.

ewortzman:
I'm seeing inconsistent behavior with importing.  When I use `import pxssh`, it works on some systems and fails on others.  I've encountered one system where `from pexpect import pxssh` fails but `import pxssh` works.  

I can't seem to find any pattern to why this happens.  All system have version 4.0.1 installed using `pip`, but some have a `pxssh.py` file installed directly to $PYTHONHOME/site-packages, while others lack this file.

takluyver:
I would guess that some of your systems have files from an older install of pexpect, or an older version installed through another mechanism. `from pexpect import pxssh` is the supported way to import it; if that fails, then I would guess it's finding an old version of pexpect; do `import pexpect; print(pexpect)` to see where it gets it from.

ewortzman:
That makes sense.  I looked at the `pxssh.py` file and all it does is issue a warning that the module has been moved then import the correct module.  I'll update my files with the correct method of importing.

PR #514, opened 2018-07-12
Add note to pxssh doc inviting people to consider Paramiko

takluyver:
cc @Red-M @jquast @jnatschev, following the discussion on #505.

Despite the issues @Red-M reported seeing, I think Paramiko is generally the better option for people who need to do things programmatically over SSH. It avoids the overhead and complexity of shelling out to ssh, and it can do more by speaking the protocol directly - e.g. it's easy to open an SFTP client on an existing connection.

I don't think there's any need to get rid of pxssh, but given that we often recommend paramiko when people have trouble with pexpect (e.g. #339, #414), it seems like a good idea to point people there in the docs.

### DIFF
--- a/doc/api/pxssh.rst
+++ b/doc/api/pxssh.rst
@@ -1,6 +1,14 @@
 pxssh - control an SSH session
 ==============================
 
+.. note::
+
+   *pxssh* is a screen-scraping wrapper around the SSH command on your system.
+   In many cases, you should consider using
+   `Paramiko <https://github.com/paramiko/paramiko>`_ instead.
+   Paramiko is a Python module which speaks the SSH protocol directly, so it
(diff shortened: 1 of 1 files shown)


Red-M:
Sorry to bring this back but I've actually built a potential replacement for pxssh which should very closely emulate pxssh with paramiko with additional functionality of paramiko_expect.

Its in a very early stages but it should replace this completely as you can request a lot of things that require hacks in pxssh to do, like opening extra ssh tunnels during the connection or SFTP interaction in its entirety (which required just pexpect calling sftp).

The library is over on my bitbucket or at https://github.com/Red-M/RedSSH

takluyver:
:+1: sounds useful. Whenever you're happy enough with it to point more people towards it, feel free to add a note about it to the pxssh docs.

Issue #521, opened 2018-08-20
"test_bash" is failed


stanislavlevin:
Some OS still use an old bash. And there is no builtin "time" command. For instance, fail case:
```sh
bash --version
GNU bash, version 3.2.57(1)-release (x86_64-alt-linux-gnu)
```
```
type time
time is hashed (/usr/bin/time)
```
```
time
Usage: time [-apvV] [-f format] [-o file] [--append] [--verbose]
       [--portability] [--format=format] [--output=file] [--version]
       [--quiet] [--help] command [arg...]
```
And of course, an error is here:
```
tests/test_expect.py:418: AssertionError
----------------------------- Captured stdout call -----------------------------

 tests.test_expect.ExpectTestCase.test_before_across_chunks 
__________________________ REPLWrapTestCase.test_bash __________________________

self = <tests.test_replwrap.REPLWrapTestCase testMethod=test_bash>

    def test_bash(self):
        bash = replwrap.bash()
        res = bash.run_command("time")
>       assert 'real' in res, res
E       AssertionError: Usage: time [-apvV] [-f format] [-o file] [--append] [--verbose]
E                [--portability] [--format=format] [--output=file] [--version]
E                [--quiet] [--help] command [arg...]
E         
\n'      [--quiet] [--help] command [arg...]--output=file] [--version]-append] [--verbose]

tests/test_replwrap.py:28: AssertionError
```

And for instance a success case:
```sh
bash --version
GNU bash, version 4.4.23(1)-release (x86_64-redhat-linux-gnu)
```
```
type time
time is a shell keyword
```
```
time

real    0m0.000s
user    0m0.000s
sys     0m0.000s
```

So, of course, results are different.

PR #522, opened 2018-08-20
Fix "test_bash" test

stanislavlevin:
Due to an old bash version(3.x) on some systems there is no builtin "time"
command. Another more common one can be used for.

Fixes: https://github.com/pexpect/pexpect/issues/521


takluyver:
Thanks

Issue #523, opened 2018-08-20
"test_before_across_chunks" is failed


stanislavlevin:
Test "test_before_across_chunks" is failed with error message:
```
___________________ ExpectTestCase.test_before_across_chunks ___________________

self = <tests.test_expect.ExpectTestCase testMethod=test_before_across_chunks>

    def test_before_across_chunks(self):
        # https://github.com/pexpect/pexpect/issues/478
        child = pexpect.spawn(
            '''/bin/bash -c "openssl rand -base64 {} | head -500 | nl --number-format=rz --number-width=5 2>&1 ; echo 'PATTERN!!!'"'''.format(1024 * 1024 * 2),
            searchwindowsize=128
        )
        child.expect(['PATTERN'])
>       assert len(child.before.splitlines()) == 500
E       AssertionError: assert 502 == 500
```

On some systems the following command output is differ from expected:
```sh
openssl rand -base64 2097152 | head -n 500
...
Y7lVPQF9lnUvt1lQeePlesLTsBTIWyZ99ARGpdbaWTsB7vpZbLpGTs/TXpbZh8SE
140409058122752:error:02012020:system library:fflush:Broken pipe:bss_file.c:434:fflush()
140409058122752:error:20074002:BIO routines:FILE_CTRL:system lib:bss_file.c:436:
```

PR #524, opened 2018-08-20
Fix "test_before_across_chunks" test

stanislavlevin:
Pipe like "openssl rand -base64 2097152 | head -n 500" leads to "Broken pipe"
error on some systems. This is an unexpected behavior. So errors must be
suppressed.

Fixes: https://github.com/pexpect/pexpect/issues/523

### DIFF
--- a/tests/test_expect.py
+++ b/tests/test_expect.py
@@ -411,7 +411,7 @@ class ExpectTestCase (PexpectTestCase.PexpectTestCase):
     def test_before_across_chunks(self):
         # https://github.com/pexpect/pexpect/issues/478
         child = pexpect.spawn(
-            '''/bin/bash -c "openssl rand -base64 {} | head -500 | nl --number-format=rz --number-width=5 2>&1 ; echo 'PATTERN!!!'"'''.format(1024 * 1024 * 2),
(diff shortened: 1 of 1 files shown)


stanislavlevin:
Hello, how can i fix coverage?
Please advise.

bmwiedemann:
IMHO the coveralls result is bogus, because it does not recognize that the change is in a test (and tests always cover themselves in addition to the code under test)

I independently discovered and submitted the same patch with this description:
```
Without this patch, the test failed with openssl-1.0.2p
because 502 lines were captured with the extra lines being at the end:

140510355240592:error:02012020:system library:fflush:Broken pipe:bss_file.c:434:fflush()
140510355240592:error:20074002:BIO routines:FILE_CTRL:system lib:bss_file.c:436:

neither openssl-1.0.2j nor 1.1.0h needed this patch.
```

I also learned this:
The changes around the fflushing were intentional:
https://github.com/openssl/openssl/pull/3348

It was a fix for:
https://github.com/openssl/openssl/issues/3232

takluyver:
Thanks, seems reasonable. I don't know why coveralls reckons the coverage decreased, but it's only by a tiny amount, so I'm just going to ignore it.

PR #527, opened 2018-09-04
tests: ignore stderr from openssl

bmwiedemann:
Without this patch, the test failed with openssl-1.0.2p
because 502 lines were captured with the extra lines being at the end:
```
140510355240592:error:02012020:system library:fflush:Broken pipe:bss_file.c:434:fflush()
140510355240592:error:20074002:BIO routines:FILE_CTRL:system lib:bss_file.c:436:
```

neither openssl-1.0.2j nor 1.1.0h needed this patch.


bmwiedemann:
duplicate of #524

PR #528, opened 2018-09-05
Allow customized pxssh client

tschmid:
This can be very useful if you want to run the pxssh client in a network namespace.

### DIFF
--- a/pexpect/pxssh.py
+++ b/pexpect/pxssh.py
@@ -259,7 +259,7 @@ class pxssh (spawn):
                 sync_multiplier=1, check_local_ip=True,
                 password_regex=r'(?i)(?:password:)|(?:passphrase for key)',
                 ssh_tunnels={}, spawn_local_ssh=True,
-                sync_original_prompt=True, ssh_config=None):
+                sync_original_prompt=True, ssh_config=None, cmd='ssh'):
         '''This logs the user into the given server.
 
         It uses
@@ -303,6 +303,10 @@ class pxssh (spawn):
         file to the client to handle itself. You may set any options you wish in here, however
         doing so will require you to post extra information that you may not want to if you
         run into issues.
+
+        Alter the ``cmd`` to change the ssh client used, or to prepend it with network
+        namespaces. For example ```cmd="ip netns exec vlan2 ssh"``` to execute the ssh in 
+        network namespace named ```vlan```.
         '''
         
         session_regex_array = ["(?i)are you sure you want to continue connecting", original_prompt, password_regex, "(?i)permission denied", "(?i)terminal type", TIMEOUT]
@@ -354,7 +358,7 @@ class pxssh (spawn):
                         if spawn_local_ssh==False:
                             tunnel = quote(str(tunnel))
                         ssh_options = ssh_options + ' -' + cmd_type + ' ' + str(tunnel)
-        cmd = "ssh %s -l %s %s" % (ssh_options, username, server)
+        cmd += " %s -l %s %s" % (ssh_options, username, server)
         if self.debug_command_string:
             return(cmd)
 


Red-M:
I'd prefer some tests with this option before accepting this.

tschmid:
Good call. How about I add a test similar to: test_ssh_tunnel_string making sure that the command string changes if you change this value?

Red-M:
I'd do it so that it tests with a different command and expects a failure
or some form of success.

Trek333:
@Red-M I am using pexpect for an app (CLI-Auto-for-Splunk), and overriding the login method of the pxssh class. I like this suggested enhancement by @tschmid . My use case might use it for appending options to the ssh command. Can you give more details on the tests that are needed here to allow acceptance of the change? I would have the same suggestion for a test case as @tschmid . If I had enough details for the test case requirements, I might submit code with other suggestions like moving the hard coded regex strings to the login method's input variables with a default value (allow users to customize the login for their ssh device's login banners) and exception code for additional ssh login failures (host key changed, max ssh sessions exceeded, etc.).

Red-M:
@trek333, At this point in time I am more inclined to recommend looking atbother libraries to get this done. I am currently making a replacement for pxssh which should allow you to do anything that pxssh can do as lo g as paramiko can connect to the device.

Trek333:
@Red-M , I tried using paramiko and another library for ssh before moving to pxssh (pexpect). I was just offering to provide my suggestions for enhancing the login method of pxssh to cover more use cases without needing to override it or create a custom login method. Based on the comments in the code of the login method, it appeared to be a work in progress.

Red-M:
I'm fine if you want to write test cases, I'm only suggesting to hold off because I'm aiming to replace pxssh completely but I've experienced issues with paramiko before and my new solution is paramiko based.

I'd prefer that such tests confirm that this feature works as intended which is to replace the ssh command with another.
I have issue with allowing for login prompt regex customisation as it will make any debugging of user issues almost impossible.
Additionally not everyone understands how regexes work 100% and as such may easily lead to extremely painful understanding of nuances of regex, what a user has done and needs to do.

Trek333:
Do you mean that the new pxssh will use paramiko instead of openssh? If so, I would request that the current pxssh version be kept and another class be written for paramiko named something like px_ssh_paramiko. We both have "experienced issues with paramiko" which is the reason I switched to pxssh (which currently uses openssh) for providing a ssh library for my app. pxssh (with openssh) fixed all of the issues that I was experiencing with ssh.  

To your reasoning of not "allowing for login prompt regex customisation" due to support reasons and pxssh users not understanding the complexities of it, I completely understand. To this point, the paramiko library would be more difficult to support and introduce significantly more complexities for users to understand also which are a couple of the reasons for switching to pxssh (with openssh) for my app. I have had no issues with pxssh (with openssh) except needing some enhancements which I added myself by overriding the login method. I was only suggesting enhancements to the login method of pxssh. It would be a shame to be stuck on the current version of pxssh due to the replacement of openssh with paramiko. If you are unaware, Splunk uses its own Python instance and not the Python instance installed for the OS. The great feature of the current pxssh is being pure Python code which allows my Splunk app to have a seamless installation. Paramiko is not pure Python code and not included on most/any Linux distros unlike openssh.

Red-M:
We're already trying to recommend that people not use pxssh.
I'm not writing a replacement as pxssh, it will be a new library.
https://github.com/pexpect/pexpect/pull/514

Trek333:
I am glad that pxssh is not being replaced. When the new library is released, I will try to take a look at it. Thanks for pointing me to the discussion about it. I am aware of others using pxssh instead of paramiko. I would suggest to not underestimate the value of pxssh especially for developers needing a pure Python solution. Also, you might check out ssh2-pyhton if you have not heard of it.

Red-M:
You'll be quite happy to know that I managed to dump paramiko in my new library for ssh2-python and it is really really fast.

Trek333:
Nice...I will try to check it out when released. However, I tried ssh2-python with my app, but as I recall, I had similar issues as paramiko due to Splunk using its own Python instance. On a related subject, I have wondered which library is better overall...ssh2-python or paramiko. In my case, I have tried both of them (and pxssh). In the end, pxssh solved all of my issues for using with my Splunk app. One concern that I had using pxssh (with openssh) was possible lower performance than ssh2-python or paramiko. For my app, pxssh's performance is fine.

Red-M:
To bring this back to the original conversation, unit tests for this are still needed to make sure that this feature works in the future.

Issue #531, opened 2018-09-17
run_command throws a TypeError exception when passed bytes


ericshawlinux:
Currently sending bytes to REPLWrapper causes a TypeError exception. I think that it should accept bytes.

```python
#!/usr/bin/python3
from pexpect.replwrap import REPLWrapper
py = REPLWrapper("python3", ">>> ", None)
py.run_command(b'\x34\x2b\x37')
```
```Python traceback
Traceback (most recent call last):
  File "./test.py", line 6, in <module>
    py.run_command(b'\x34\x2b\x37')
  File "/usr/lib/python3/dist-packages/pexpect/replwrap.py", line 82, in run_command
    if command.endswith('\n'):
TypeError: endswith first arg must be bytes or a tuple of bytes, not str
```

ericshawlinux:
I'm going to attempt to fix this and open a pull request.

PR #532, opened 2018-09-18
Fixed TypeError exception when sending bytes to REPLWrapper

ericshawlinux:
This is a fix for issue #531, which was opened by me just yesterday.

I added a check for type, to the newline check on `command`, a format string when `command` is used in an exception message below, and a test method for the new functionality, which I copied and changed a little from `test_multiline` in the same file.

### DIFF
--- a/pexpect/_async.py
+++ b/pexpect/_async.py
@@ -46,7 +46,9 @@ def repl_run_command_async(repl, cmdlines, timeout=-1):
         repl.child.kill(signal.SIGINT)
         yield from repl._expect_prompt(timeout=1, async_=True)
         raise ValueError("Continuation prompt found - input was incomplete:")
-    return u''.join(res + [repl.child.before])
+    res = res + [repl.child.before]
+    sep = u'' if isinstance(res[0], str) else b''
+    return sep.join(res)
 
 class PatternWaiter(asyncio.Protocol):
     transport = None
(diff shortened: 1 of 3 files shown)


ericshawlinux:
I will try to fix those errors from the other versions. I didn't notice them because I only tried to test python 2.7.

ericshawlinux:
I think I need advice on that test_async.py failing. a `ValueError` is expected but instead, a `pexpect.exceptions.TIMEOUT` is raised.

ericshawlinux:
This probably needs to be more robust. I think with other encodings besides None, it doesn't work.

takluyver:
The general pattern for Pexpect is meant to be that if you use `encoding=None`,  you use bytes to interact with it, and if you specify an encoding, you interact with it in unicode. If you create a REPLWrapper by passing a command, it uses `encoding='utf-8'` (since it's a relatively new, higher level interface).

If you want to use REPLWrapper with bytes, I'd say it should be by passing in a spawn object created with `encoding=None`:

```python
py = REPLWrapper(spawn("python3", encoding=None), ">>> ", None)
```

This won't work at the moment - it still runs into the bug that you reported. It could be fixed. But I partly wonder if it's better to say that `replwrap` is a unicode-only interface, and if you want to work with bytes, you should use the lower-level `spawn()` interface directly.

Issue #517, opened 2018-08-03
Receive bytes intead of str in `output_filter` under unicode encoding.


giovanism:
Based on this doc, I wrote my output_filter to match certain patterns.

> You may pass in optional input and output filter functions. These
> functions should take a string and return a string. The output_filter
> will be passed all the output from the child process. The input_filter
> will be passed all the keyboard input from the user. The input_filter
> is run BEFORE the check for the escape_character.

But this is not the case when I run it under python3.

I have to wrap around my output_filter using something like:
```
def output_filter_utf8(out_raw):
    return output_filter(out_raw.decode('utf-8')).encode('utf-8')
```

It doesn't matter whether I pass `encoding='utf-8'` or not to `spawn`.

My error message:
```
Traceback (most recent call last):
  File "/usr/lib64/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib64/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/giovanism/Repos/garbage/pys/tsudo/tsudo/__main__.py", line 4, in <module>
    tsudo.main(sys.argv[1:])
  File "/home/giovanism/Repos/garbage/pys/tsudo/tsudo/tsudo.py", line 15, in main
    sudo.interact(output_filter=insults_filter)
  File "/home/giovanism/.local/lib/python3.6/site-packages/pexpect/pty_spawn.py", line 766, in interact
    self.__interact_copy(escape_character, input_filter, output_filter)
  File "/home/giovanism/.local/lib/python3.6/site-packages/pexpect/pty_spawn.py", line 810, in __interact_copy
    data = output_filter(data)
  File "/home/giovanism/Repos/garbage/pys/tsudo/tsudo/tsudo.py", line 21, in insults_filter
    if sudo_out == '\r\n' or attempts_up_pat.match(sudo_out):
TypeError: cannot use a string pattern on a bytes-like object
```

takluyver:
The doc probably hasn't changed since Python 3 support was added. Do you want to update it to say talk about bytes rather than strings?

PR #537, opened 2018-10-08
interact: Add Python 3 compatibility warning

giovanism:
Close https://github.com/pexpect/pexpect/issues/517

### DIFF
--- a/pexpect/pty_spawn.py
+++ b/pexpect/pty_spawn.py
@@ -752,10 +752,14 @@ class spawn(SpawnBase):
         child process in interact mode is duplicated to the given log.
 
         You may pass in optional input and output filter functions. These
-        functions should take a string and return a string. The output_filter
-        will be passed all the output from the child process. The input_filter
-        will be passed all the keyboard input from the user. The input_filter
-        is run BEFORE the check for the escape_character.
+        functions should take bytes array and return bytes array too. Even
+        with ``encoding='utf-8'`` support, meth:`interact` will always pass
+        input_filter and output_filter bytes. You may need to wrap your
+        function to decode and encode back to UTF-8.
+
+        The output_filter will be passed all the output from the child process.
+        The input_filter will be passed all the keyboard input from the user.
+        The input_filter is run BEFORE the check for the escape_character.
 
         Note that if you change the window size of the parent the SIGWINCH
         signal will not be passed through to the child. If you want the child
@@ -775,7 +779,7 @@ class spawn(SpawnBase):
             signal.signal(signal.SIGWINCH, sigwinch_passthrough)
             p.interact()
         '''
-
+        
         # Flush the buffer.
         self.write_to_stdout(self.buffer)
         self.stdout.flush()


pexpect/pty_spawn.py
```diff
@@ -774,6 +774,11 @@ def sigwinch_passthrough (sig, data):
             p = pexpect.spawn('/bin/bash')
             signal.signal(signal.SIGWINCH, sigwinch_passthrough)
             p.interact()
+        
+        .. note ::
+            Even after Python 3 ``unicode='utf-8'`` support, meth:`interact`
+            doesn't guaranteed to pass input_filter and output_filter str value.
+            You may need to wrap your function to decode and encode back to UTF-8.
```
takluyver: We can be more definite about this: the type is consistent: always bytes, regardless of Python version or `encoding=` argument.

I'd also like the docstring to be improved by updating the paragraph above that talks about filter functions, rather than sticking an extra note on.

giovanism:
Sorry to take me this long, how is it now? @takluyver

pexpect/pty_spawn.py
```diff
@@ -752,10 +752,14 @@ def interact(self, escape_character=chr(29),
         child process in interact mode is duplicated to the given log.
 
         You may pass in optional input and output filter functions. These
-        functions should take a string and return a string. The output_filter
-        will be passed all the output from the child process. The input_filter
-        will be passed all the keyboard input from the user. The input_filter
-        is run BEFORE the check for the escape_character.
+        functions should take bytes array and return bytes array too. Even
+        after Python 3 ``unicode='utf-8'`` support, meth:`interact` doesn't
```
takluyver: Do you mean `encoding='utf-8'`?

Also, this is not specific to Python 3; the same is true in Python 2 (with `unicode` in place of `str`).

pexpect/pty_spawn.py
```diff
@@ -752,10 +752,14 @@ def interact(self, escape_character=chr(29),
...
+        guarantee to pass input_filter and output_filter encoded str value.
```
takluyver: "doesn't guarantee" sounds like it can vary, which isn't the case: it always passes bytes.

Also, "encoded str" isn't quite right. A Python 3 str is *decoded*. Decoding converts bytes to str, encoding converts str to bytes.

giovanism: Will fix

pexpect/pty_spawn.py
giovanism: Will fix

giovanism:
I should have write more commits and squashed only at last. Anyway here's the fix

pexpect/pty_spawn.py
```diff
@@ -752,10 +752,14 @@ def interact(self, escape_character=chr(29),
         child process in interact mode is duplicated to the given log.
 
         You may pass in optional input and output filter functions. These
-        functions should take a string and return a string. The output_filter
-        will be passed all the output from the child process. The input_filter
-        will be passed all the keyboard input from the user. The input_filter
-        is run BEFORE the check for the escape_character.
+        functions should take bytes array and return bytes array too. Even
+        with ``encoding='utf-8'`` support, meth:`interact` will always pass
+        input_filter and output_filter decoded value. You may need to wrap your
```
takluyver: Again, these are not 'decoded values'.

- Decode: bytes -> str
- Encode: str -> bytes

giovanism: Sorry, i just skimed through your comment and went to switch (de/en)code.

So, can we say that this "_encoded_ value" get to passed to filters? And, in python3, filters do pass bytes, thus i had thought that "encoded str" is the best term. How should we express this in python2/3 compatible way?

giovanism: @takluyver

takluyver: I'd just say "will always pass input_filter and output_filter bytes".

`bytes` is meaningful on Python 2 and 3 - on Python 3 it's the name of the separate bytes type, and on Python 2 it's an alias for `str`. So I think it's fine to use that name.

giovanism:
@takluyver

takluyver:
Thanks

PR #539, opened 2018-10-17
Protect log writer from encoded bytestrings

jpic:


quintenroets (commented):
This solves my issue

Red-M:
Due to the number of failures in the tests (and what the failures are) when this ran, I don't think this is a good idea to merge as is.

quintenroets:
Agree.
The workaround that I use now is:
`
import pexpect
import tempfile

with tempfile.TemporaryFile() as tmp:
    child = pexpect.spawn(command, timeout=None, logfile=tmp)
    child.interact()
    tmp.seek(0)
    logs = tmp.read().decode()
`

quintenroets:
Agree.
The workaround I now use:

import pexpect
import tempfile

with tempfile.TemporaryFile() as tmp:
    child = pexpect.spawn(command, timeout=None, logfile=tmp)
    child.interact()
    tmp.seek(0)
    logs = tmp.read().decode()

quintenroets:
Agree.
The workaround I now use:

import pexpect
import tempfile

with tempfile.TemporaryFile() as tmp:
    child = pexpect.spawn(command, timeout=None, logfile=tmp)
    child.interact()
    tmp.seek(0)
    logs = tmp.read().decode()

quintenroets (commented):
The code does not check if the self.encoding is indeed utf-8.
The variable s only needs to be decoded if self.encoding is different than None. And the argument of decode should be self.encoding

PR #556, opened 2019-02-05
make sure we're dealing with strings

kcr:
This seems to be necessary to make .interact() work under certain circumstances.


kcr:
*headdesk* or maybe I was using it wrong.

takluyver:
That doesn't look quite right. The log files should get bytes data if the spawn object doesn't have a specified encoding, and unicode data if it does ([docs](https://pexpect.readthedocs.io/en/stable/api/pexpect.html#pexpect.spawn.logfile)).

It looks like the interact machinery is sending bytes to the log files regardless of whether there's a specified encoding. You don't need to decode the data to pass it between stdin and the child terminal, but that introduces an inconsistency in the way the data is logged.

Issue #558, opened 2019-03-01
Expect raises `TypeError` in Python 3 if patterns include RegExp compiled from a non-bytes type string


EmersonPrado:
> From commit `03168dd3cb57cbafebccef0d46e00c6bbd4bb0b9`
> Python 3.6.7-1~18.04 in Ubuntu 18.04

Docs say Pexpect accepts compiled regular expressions as patterns:

```Shell
pydoc3 pexpect.spawn.compile_pattern_list	# Idem for pexpect.spawnbase
"Patterns must be a StringType, EOF, TIMEOUT, SRE_Pattern, or a list of those"
pydoc3 pexpect.spawn.expect			# Idem for pexpect.spawnbase
"The pattern can be a StringType, EOF, a compiled re, or a list of any of those types"
pydoc3 pexpect.spawn.expect_list		# Idem for pexpect.spawnbase
"This takes a list of compiled regular expressions"
pydoc3 pexpect.expect.searcher_re.__init__
"'patterns' may be a list or other sequence of compiled regular expressions"
```

But, if we do provide a pattern with compiled RegExp, class `expect.search` raises exception `TypeError` in line 294 (`match = s.search(buffer, searchstart)`)

```Python
>>> from re import compile as re_compile
>>> from lib.pexpect.pexpect import spawn, EOF, TIMEOUT
>>>
>>> spawn_object = spawn('ls')
>>> spawn_object.expect([re_compile(': '), EOF, TIMEOUT])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<Pexpect clone>/pexpect/spawnbase.py", line 341, in expect
    timeout, searchwindowsize, async_)
  File "<Pexpect clone>/pexpect/spawnbase.py", line 369, in expect_list
    return exp.expect_loop(timeout)
  File "<Pexpect clone>/pexpect/expect.py", line 103, in expect_loop
    idx = self.new_data(incoming)
  File "<Pexpect clone>/pexpect/expect.py", line 29, in new_data
    index = searcher.search(window, len(data))
  File "<Pexpect clone>/pexpect/expect.py", line 294, in search
    match = s.search(buffer, searchstart)
TypeError: cannot use a string pattern on a bytes-like object
>>>
```

Pexpect parses and searches correctly a string pattern, or a RegExp pattern compiled from a bytes-type string

```Python
>>> from re import compile as re_compile
>>> from lib.pexpect.pexpect import spawn, EOF, TIMEOUT
>>>
[...]

PR #560, opened 2019-03-05
[Issue #558] Coerce compiled regex patterns type according to spawn encoding

EmersonPrado:
This change fixes [issue 558](https://github.com/pexpect/pexpect/issues/558) by extending the type coercing already present in exact matches for compiled regex matches. In short: if spawn encoding is None, it changes the compiled regex to have a bytes type pattern. Otherwise, changes to non-bytes type pattern.


Red-M:
Thanks!

PR #562, opened 2019-03-21
pxssh.login require either username or ssh_config

tgbugs:
If pxssh.login is called with an ssh_config then do not require username since it can be supplied via the config and some programs do not want to manage the username at all since it would require parsing the ssh config.

This commit removes the independent username parameter when formatting cmd and instead appends the flag and the username to ssh_options if a username is provided.

If neither username nor ssh_config is provided raise a TypeError in line with the error raised natively if username was not provided as a required positional argument.

### DIFF
--- a/pexpect/pxssh.py
+++ b/pexpect/pxssh.py
@@ -253,7 +253,7 @@ class pxssh (spawn):
     ### TODO: This is getting messy and I'm pretty sure this isn't perfect.
     ### TODO: I need to draw a flow chart for this.
     ### TODO: Unit tests for SSH tunnels, remote SSH command exec, disabling original prompt sync
-    def login (self, server, username, password='', terminal_type='ansi',
+    def login (self, server, username=None, password='', terminal_type='ansi',
                 original_prompt=r"[#$]", login_timeout=10, port=None,
                 auto_prompt_reset=True, ssh_key=None, quiet=True,
                 sync_multiplier=1, check_local_ip=True,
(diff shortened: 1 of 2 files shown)


Red-M:
My concern with this is parsing the config to confirm that the username string is present, otherwise the current user's username will be used which could be very bad. Please have this confirm that the config contains a username and include this in the unit tests.

tgbugs:
Thank you for catching that. Boneheaded of me. I think the tests that I added cover all the cases where the user's local name might leak out.

Red-M:
Do you mind rebasing this change?

tgbugs:
Looks like the failures are all caused by tests that were added by 978c58f4 which this was never check against. @Trek333 would you mind taking a look?

Red-M:
I think that this is the inter-operation between these two PRs, I'll fix this soon.

pexpect/pxssh.py
```diff
@@ -358,7 +358,43 @@ def login (self, server, username, password='', terminal_type='ansi',
                         if spawn_local_ssh==False:
                             tunnel = quote(str(tunnel))
                         ssh_options = ssh_options + ' -' + cmd_type + ' ' + str(tunnel)
-        cmd += " %s -l %s %s" % (ssh_options, username, server)
```
Red-M: This line needs to be added back, the rebase removed it.

tgbugs: I misread that and removed it in the merge from master into patch-1 because cmd is overwritten on 396 (I think). The custom ssh command option looks like it needs to be put in at 396 if I'm not mistaken.

pexpect/pxssh.py
```diff
@@ -358,7 +358,43 @@ def login (self, server, username, password='', terminal_type='ansi',
...
+                del line
+
+            del lines
+
+            if not config_has_server:
+                raise TypeError('login() ssh_config has no Host entry for %s' % server)
+            elif not server_has_username:
+                raise TypeError('login() ssh_config has no user entry for %s' % server)
+
+        cmd = "ssh %s %s" % (ssh_options, server)
```
tgbugs: I think this is the fix.
```python
cmd += " %s %s" % (ssh_options, server)
```

tgbugs:
Ok, that force push should have resolved the issue properly.

Red-M:
CI looks good!

Trek333:
@Red-M and @tgbugs  - Strange...I get the error below for the "test_fake_ssh" test. It appears that self.before variable is not returning the prompt strings of the ssh session which raises the exception. Also, the "test_custom_ssh_cmd" test is failing for the same reason. Can you verify that the "test_fake_ssh" test is not failing for you?
_____________________________ PxsshTestCase.test_fake_ssh _____________________________
...
        if auto_prompt_reset:
            if not self.set_unique_prompt():
                self.close()
                raise ExceptionPxssh('could not set shell prompt '
                                     '(received: %r, expected: %r).' % (
>                                        self.before, self.PROMPT,))
E               ExceptionPxssh: could not set shell prompt (received: "unset PROMPT_COMMAND\r\nPS1='[PEXPECT]\\$ '\r\nset prompt='[PEXPECT]\\$ '\r\n", expected: '\\[PEXPECT\\][\\$\\#] ').

Red-M:
I'm not getting any errors, just make sure you've installed pexpect for the version of python your using from the git repo rather than pypi.

Trek333:
@Red-M  Thanks for the response. The git repo was downloaded again which resolved the errors from test scripts.

PR #574, opened 2019-07-04
Docs: distinguish class attributes and object attributes

xtofl:
When reading the documentation of the 'commonissues', I was a little confused:
the documentation mentioned 'the delaybeforesend attribute of the spawn class'.

Since classes are actual things that can have attributes, it contradicts the example
of two lines below.

This MR rectifies this small inconsistency.


takluyver:
Seems reasonable, thanks.

PR #575, opened 2019-07-07
Add doc note about new expect based SSH automation library

Red-M:
As per #514, I'm finally happy with this library enough to call it stable.

It needs a little work around unit tests and I probably  need to clean the code a little bit but the code is mature enough that someone other than me could use it as I'm not finding bugs in the library code and more bugs in what ever I've hacked up to use it.

### DIFF
--- a/doc/api/pxssh.rst
+++ b/doc/api/pxssh.rst
@@ -5,9 +5,16 @@ pxssh - control an SSH session
 
    *pxssh* is a screen-scraping wrapper around the SSH command on your system.
    In many cases, you should consider using
-   `Paramiko <https://github.com/paramiko/paramiko>`_ instead.
+   `Paramiko <https://github.com/paramiko/paramiko>`_ or
+   `RedExpect <https://github.com/Red-M/RedExpect>`_ instead.
    Paramiko is a Python module which speaks the SSH protocol directly, so it
    doesn't have the extra complexity of running a local subprocess.
+   RedExpect is very similar to pxssh except that it reads and writes directly
+   into an SSH session all done via python with all the SSH protocol in C,
+   addtionally it is written for communicating to SSH servers that are not just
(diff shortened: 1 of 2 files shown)


doc/api/pxssh.rst
```diff
@@ -5,9 +5,16 @@ pxssh - control an SSH session
...
-   `Paramiko <https://github.com/paramiko/paramiko>`_ instead.
+   `Paramiko <https://github.com/paramiko/paramiko>`_ or
+   `RedExpect <https://github.com/Red-M/RedExpect>`_ instead.
    Paramiko is a Python module which speaks the SSH protocol directly, so it
    doesn't have the extra complexity of running a local subprocess.
+   RedExpect is very similar to pxssh except that it reads and writes directly
+   into an SSH session all done via python with all the SSH protocol in C,
+   addtionally it is written for communicating to SSH servers that are not just
+   linux machines. Meaning that it is extremely fast in comparison to Paramiko
+   and already has the expect enviroment that you expect to have. In most cases
```
takluyver: Do you mean 'API' instead of 'environment' here?

Also, maybe it's best to avoid using 'expect' with two different meanings, e.g. by saying "the familiar expect API"

takluyver:
Nice, thanks. I've suggested a couple of changes to the wording, but feel free to merge this whenever you're happy with it.

takluyver:
Nice, thanks. I've suggested a couple of changes to the wording, but feel free to merge this whenever you're happy with it.

doc/api/pxssh.rst
Red-M: It was more a word joke but I'll change that.

takluyver: I did actually smile at it, but I'm a native English speaker with a fondness for wordplay. :-)

PR #578, opened 2019-08-01
Fix missing data in 'before', especially after timeouts

dluyer:
- Fix 'before' return on eof/timeout/error.
- Fix behavior when searchwindowsize is increased between calls to expect_loop (existing code could discard required content).

Note, this is less efficient, but it is correct.  The prior changes improved benchmarks but broke correctness; I haven't benchmarked to see if this is better or worse than the original code, before buffer_type use was introduced.  It would be possible to make this more efficient with further complexity.

Breakage introduced:
  https://github.com/pexpect/pexpect/commit/fd7332f59e003a8ee658647f883d4b03018c5e5b#diff-244f4b51bfbbbcb072d1428a79b9f4f7

Prior partial fix:
  https://github.com/pexpect/pexpect/commit/5a2e63f3fbfb39458c832a65bf8db330e954b8f6#diff-244f4b51bfbbbcb072d1428a79b9f4f7

### DIFF
--- a/pexpect/expect.py
+++ b/pexpect/expect.py
@@ -32,19 +32,20 @@ class Expecter(object):
             spawn._buffer.write(window[searcher.end:])
             spawn.before = spawn._before.getvalue()[0:-(len(window) - searcher.start)]
             spawn._before = spawn.buffer_type()
-            spawn.after = window[searcher.start: searcher.end]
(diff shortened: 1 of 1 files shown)


dluyer:
Note, obsoleted by #579 (this change plus efficiency improvements).

PR #580, opened 2019-08-05
Always use self.PYTHONBIN not 'python' or sys.executable in tests

dluyer:
Support cases where self.PYTHONBIN != sys.executable, by consistently using self.PYTHONBIN in tests.  Then, anyone who has a problematic value for sys.executable only needs to edit PexpectTestCase in a local patch (much smaller local patch to maintain).

### DIFF
--- a/tests/test_interact.py
+++ b/tests/test_interact.py
@@ -42,8 +42,6 @@ class InteractTestCase (PexpectTestCase.PexpectTestCase):
             env['PYTHONPATH'] = self.project_dir
 
         self.interact_py = ('{self.PYTHONBIN} interact.py'.format(self=self))
-        if self.PYTHONBIN != sys.executable:
-            self.interact_py += ' --pythonbin {self.PYTHONBIN}'.format(self=self)
 
     def test_interact_escape(self):
         " Ensure `escape_character' value exits interactive mode. "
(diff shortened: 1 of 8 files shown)


dluyer:
Added fix for test_ctrl_chars, where it was simply running 'python' not even sys.executable.

tests/test_interact.py
```diff
@@ -41,7 +41,9 @@ def setUp(self):
         else:
             env['PYTHONPATH'] = self.project_dir
 
-        self.interact_py = ('{sys.executable} interact.py'.format(sys=sys))
+        self.interact_py = ('{self.PYTHONBIN} interact.py'.format(self=self))
+        if self.PYTHONBIN != sys.executable:
+            self.interact_py += ' --pythonbin {self.PYTHONBIN}'.format(self=self)
```
takluyver: I'm unsure why this bit is needed - surely if you run the subprocess with `self.PYTHONBIN`, then its `sys.executable` will always be the Python you want?

dluyer: That's possible, actually, I'll test without this part of this change.

dluyer: Removed this and the related unnecessary change in interact.py from the pull request.

PR #581, opened 2019-08-05
Use IPv6 loopback if there is no IPv4 loopback available

dluyer:
Fix test failures on IPv6-only machines.


takluyver:
Looks reasonable.

PR #582, opened 2019-08-06
Fix test_ansi.py when running in a non-writable directory

dluyer:
Minor fix, test_ansi causes ANSI.py's DoLog to write to cwd.

If the test directory is not writable, this fails.  It should run from a tempdir.


takluyver:
I think you've try/finally-ed the wrong bit - the code you've surrounded doesn't do anything with ANSI.

Other than that, this looks like a reasonable workaround. But it would probably be better overall if the logging could be set not to write to the CWD.

dluyer:
Corrected, sorry.  I'm working in an internal source control system and have been uploading via copy/paste through the github UI, and wrapped the wrong lines.

PR #584, opened 2019-08-06
Fix typo in one test name, and rename another with no valid tests

dluyer:
'depricated' should be 'deprecated'; also, test_run_out_of_pty is entirely disabled, so similarly rename it to deprecated.


takluyver:
Thanks

PR #590, opened 2019-09-25
fix select func override select module

danielino:
There is a rare circumstance when select module is replaced or disturbed by select() function in pty_spawn.py

```
  File "/Users/****/****/venv/lib/python3.7/site-packages/pexpect/pty_spawn.py", line 446, in select
    return poll_ignore_interrupts([self.child_fd], timeout)
  File "/Users/****/****/venv/lib/python3.7/site-packages/pexpect/utils.py", line 166, in poll_ignore_interrupts
    poller = select.poll()
AttributeError: module 'select' has no attribute 'poll'

```

debugging  with breakpoint() in select.utils module

```
> /Users/****/*****/venv/lib/python3.7/site-packages/pexpect/utils.py(168)poll_ignore_interrupts()
-> poller = select.poll()
(Pdb) dir(select)
[ ### ....  A LOT OF  CONSTANT ...... ###, 'POLLWRNORM', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', 'error', 'select'] <-- some methods are missing from module.
````
replacing function in pty_spawn.py fix the issue.

```
(Pdb) dir(select)
[ ....... ', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', 'error', 'kevent', 'kqueue', 'poll', 'select']
```

Signed-off-by: Daniele Marcocci <daniele.marcocci@par-tec.it>

### DIFF
--- a/pexpect/pty_spawn.py
+++ b/pexpect/pty_spawn.py
@@ -442,10 +442,10 @@ class spawn(SpawnBase):
             raise ValueError('I/O operation on closed file.')
 
         if self.use_poll:
-            def select(timeout):
+            def _select(timeout):
                 return poll_ignore_interrupts([self.child_fd], timeout)
         else:
-            def select(timeout):
+            def _select(timeout):
                 return select_ignore_interrupts([self.child_fd], [], [], timeout)[0]
 
         # If there is data available to read right now, read as much as
@@ -454,14 +454,14 @@ class spawn(SpawnBase):
         # often. See also:
         # * https://github.com/pexpect/pexpect/pull/304
         # * http://trac.sagemath.org/ticket/10295
-        if select(0):
+        if _select(0):
             try:
                 incoming = super(spawn, self).read_nonblocking(size)
             except EOF:
                 # Maybe the child is dead: update some attributes in that case
                 self.isalive()
                 raise
-            while len(incoming) < size and select(0):
+            while len(incoming) < size and _select(0):
                 try:
                     incoming += super(spawn, self).read_nonblocking(size - len(incoming))
                 except EOF:
@@ -481,7 +481,7 @@ class spawn(SpawnBase):
             # still try to read from the child_fd -- it will block
             # forever or until TIMEOUT. For that reason, it's important
             # to do this check before calling select() with timeout.
-            if select(0):
+            if _select(0):
                 return super(spawn, self).read_nonblocking(size)
             self.flag_eof = True
             raise EOF('End Of File (EOF). Braindead platform.')
@@ -496,7 +496,7 @@ class spawn(SpawnBase):
         # Because of the select(0) check above, we know that no data
         # is available right now. But if a non-zero timeout is given
         # (possibly timeout=None), we call select() with a timeout.
-        if (timeout != 0) and select(timeout):
+        if (timeout != 0) and _select(timeout):
             return super(spawn, self).read_nonblocking(size)
 
         if not self.isalive():

### CODE REVIEW

**Comment by takluyver** [comment]
That doesn't seem right. Defining a function inside another function wouldn't affect the namespace of a separate module (unless there's some horrible and unlikely bug in Python itself). And if the select function was replacing the select module somehow, the error message would say `'function' object has no attribute 'poll'` (instead of `module 'select'...`). It also doesn't have that set of attributes.

Sorry, but I think that the issue coincidentally went away at the same time you made the change, for some other reason.

There are some platforms where poll is not supported, according to the Python docs. I'm not sure which those are, but if you've got one, you'd presumably see the problem all the time.

**Comment by danielino** [comment]
i'm using osx, but poll is working well:

```
Python 3.7.4 (default, Jul  9 2019, 18:13:23) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.2.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import select                                                                                                                                                                                                                     

In [2]: select.poll()                                                                                                                                                                                                                     
Out[2]: <select.poll at 0x10764e2a0>

In [3]:  
```

this is the full stack trace 
```
Exception in thread Thread-2:
  File "/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/lib/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/danielino/AnsibleWorker/venv/lib/python3.7/site-packages/ansible_runner/runner.py", line 177, in run
    searchwindowsize=100)
  File "/Users/danielino/AnsibleWorker/venv/lib/python3.7/site-packages/pexpect/spawnbase.py", line 341, in expect
    timeout, searchwindowsize, async_)
  File "/Users/danielino/AnsibleWorker/venv/lib/python3.7/site-packages/pexpect/spawnbase.py", line 369, in expect_list
    return exp.expect_loop(timeout)
  File "/Users/danielino/AnsibleWorker/venv/lib/python3.7/site-packages/pexpect/expect.py", line 111, in expect_loop
    incoming = spawn.read_nonblocking(spawn.maxread, timeout)
  File "/Users/danielino/AnsibleWorker/venv/lib/python3.7/site-packages/pexpect/pty_spawn.py", line 457, in read_nonblocking
    if myselect(0):
  File "/Users/danielino/AnsibleWorker/venv/lib/python3.7/site-packages/pexpect/pty_spawn.py", line 446, in myselect
    return poll_ignore_interrupts([self.child_fd], timeout)
  File "/Users/danielino/AnsibleWorker/venv/lib/python3.7/site-packages/pexpect/utils.py", line 167, in poll_ignore_interrupts
    poller = select.poll()
AttributeError: module 'select' has no attribute 'poll'

```

i try to restore the file and the issue go back.

takluyver:
It looks like the select module has some... *interesting* code which runs on Macs, trying to detect a broken poll implementation. If it finds one, it deletes the poll attribute from the module:
https://github.com/python/cpython/blob/1dc1acbd73f05f14c974b7ce1041787d7abef31e/Modules/selectmodule.c#L2467-L2474

I can see from your paths that you're on Mac, so I suspect you're hitting some kind of race condition with this, so it occasionally decides that poll is broken and deletes it from the module. The apparent effect of changing code in pexpect might be random chance, or something (maybe bytecode compilation) may be affecting the race condition enough to make it more or less likely.

I have been [nerd sniped](https://www.xkcd.com/356/) :wink: .

takluyver:
If you can boil it down to a minimal example which reproduces the issue, even if it has to be launched in a shell loop until it fails, please do open a Python issue at https://bugs.python.org/

Issue #566, opened 2019-04-10
The login_timeout option make no sense in pxssh.login


BaijingML:
when I tried to login an unreachable  server: 
connect = pxssh.pxssh()
connect.login(hostname, username, password, login_timeout=6000)
as I changed the range of login_timeout from 10 to 6000, it seems nothing changed, always cost 3.1326s on my computer

rustyhowell:
@BaijingML This is because the underlying ssh process has terminated with a "No route to host" error probably. The login_timeout option is used to tell pexpect to kill the underlying ssh process if it still hasn't logged in or died when the timeout expires.

rustyhowell:
The trickier part with login_timeout is when the host exists, and the ssh command is retrying (it has it's own timeout option).  If the ssh process is still trying to login and the login_timeout expires, pxssh.login() seems to just assume that things are good and returns True.

This has caused me problems this week when debugging code that was trying to connect to a device that had just been rebooted and wasn't fully up yet. The ssh process was still trying to connect, but pxssh.login() finished anyway.  
Then I get weird EOF or TIMEOUT errors.

rustyhowell:
Proposed merge request #594

PR #594, opened 2019-10-04
Make login_timeout functional

rustyhowell:
The ssh terminal has it's own timeout option called 'CommandTimeout'. There are
scenarios where the ssh command is still trying to connect when pxssh times out.

In this case, pxssh assumes a successfull connection and returns True, even though
no authentication has happened yet. Subsequent calls to sendline and expect()
might get EOF or TIMEOUT errors. This leaves the user confused why an EOF occured
after the login() completed successfully.

### DIFF
--- a/pexpect/pxssh.py
+++ b/pexpect/pxssh.py
@@ -313,6 +313,7 @@ class pxssh (spawn):
         session_init_regex_array.extend(session_regex_array)
         session_init_regex_array.extend(["(?i)connection closed by remote host", EOF])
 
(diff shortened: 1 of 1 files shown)


rustyhowell:
This addresses #566

Red-M:
An issue with this is if a float is used for `login_timeout` or if `login_timeout` is `1` or `login_timeout=<0`, additionally tests will need to be added to make sure that the functionality will continue to work in the long term.

Issue #587, opened 2019-08-18
Example of using pexpect with pyte


dbarnett:
For curses programs, the pexpect FAQ says

> We suggest using the pyte library to screen-scrape.

Is there a simple example of how to do that, say how to start a vim instance and read out the characters on screen?

eldipa:
I don't have a "simple" example but I can show you how I use ``pexpect`` and ``pyte`` in [byexample](https://byexamples.github.io). I hope that you have a much clearer idea of how those two libs work together after this.

First, I [create](https://github.com/byexamples/byexample/blob/251a37f3119d4c3abce124a3b4a3ab2866466e05/byexample/runner.py#L162-L166) a ``Stream`` and a ``Screen``:
```python
    def _create_terminal(self, options):
        rows, cols = options['geometry']

        self._screen = Screen(cols, rows)
        self._stream = Stream(self._screen)
```

Then, I [run](https://github.com/byexamples/byexample/blob/251a37f3119d4c3abce124a3b4a3ab2866466e05/byexample/runner.py#L90-L100)  ``pexpect.spawn`` as usual.
```python
        rows, cols = options['geometry']
        self._terminal_default_geometry = (rows, cols)

        env = os.environ.copy()
        env.update({'LINES': str(rows), 'COLUMNS': str(cols)})

        self.interpreter = pexpect.spawn(cmd, echo=False,
                                                encoding=self.encoding,
                                                dimensions=(rows, cols),
                                                env=env)
```

In both cases I configure the same terminal dimension (rows and columns). I found that some programs honor the environment variables ``LINES`` and ``COLUMNS`` while others honor the 
configuration of ``pexpect`` (If you track this you will see that it uses magical ``TIOCSWINSZ`` setting [here](https://github.com/pexpect/ptyprocess/blob/3931cd45db50ee8533b8b0fef424b8d75f7ba1c2/ptyprocess/ptyprocess.py#L118-L126))

Now, you [send data](https://github.com/byexamples/byexample/blob/251a37f3119d4c3abce124a3b4a3ab2866466e05/byexample/runner.py#L151)  to the process as usual con ``pexpect``:
```python
            self.interpreter.sendline(line)
```

And you can [read the raw data](https://github.com/byexamples/byexample/blob/251a37f3119d4c3abce124a3b4a3ab2866466e05/byexample/runner.py#L252-L256) as well:
```python
        expect = [prompt_re, pexpect.TIMEOUT]
[...]

PR #604, opened 2019-12-08
Examples of how to use Pexpect and Pyte (issue #587)

eldipa:
Example of how to use Pexpect and Pyte. See #587 .


Red-M:
Thanks!

PR #612, opened 2019-12-23
ARM64: Add ARM64 jobs in Travis-CI

ossdev07:
Travis-CI has added support for ARM64. Added ARM64 jobs in Travis-CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,16 +5,28 @@ python:
   - 3.4
   - 3.5
   - 3.6
-  - pypy
-  - nightly
-
+arch:
+  - amd64
+  - arm64
 matrix:
   include:
   - python: 3.7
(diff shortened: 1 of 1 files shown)


Red-M:
Thanks for the PR but I'm not seeing the advantage we get from this.
Pexpect doesn't have any real reliance on what CPU type its running on.

rhenwood-arm:
@Red-M: I have been working with @ossdev07 on a down-stream project on AArch64 (ARM64) that has pexpect as a dependency. pexpect is a pure python project with a straightforward travis-ci build job, it doesn't have a binary release, nor does it do things that are timing sensitive. In this case, as you comment, there is apparently very little benefit in testing on alternative architectures, and this PR can probably be held in reserve until it is needed.

Red-M:
Pexpect doesn't need this as we don't depend on anything to do with the architecture of the CPU.

PR #620, opened 2020-01-22
Add project urls to `setup.py`

jugmac00:
- bug tracker
- documentation
- source code

These will be shown at the project's PyPi page ( https://pypi.org/project/pexpect/ ) and make it easier to find relevant information.


takluyver:
Thanks

PR #621, opened 2020-01-23
'host' in line not working on conf lines with: StrictHostKeyChecking,…

myururdurmaz:
current code cant find user on this entry
```
    StrictHostKeyChecking no
    UserKnownHostsFile /dev/null
    HostName 127.0.0.1
    Port 2221
    User root
    Compression yes
```


Red-M:
Could you please add tests to make sure that this bug will not be reintroduced and that [code coverage doesn't fall](https://coveralls.io/builds/28276879/source?filename=pexpect%2Fpxssh.py#L379)?

PR #623, opened 2020-02-06
Update doc: Wexpect is maintained

raczben:
Wexpect is a maintained (almost for a year) alternative of pexpect for Windows.
 - All (non-Linux specific) pexpect testcases passes with
 - ~80% coverage

### DIFF
--- a/doc/overview.rst
+++ b/doc/overview.rst
@@ -264,6 +264,8 @@ When run by ``PopenSpawn``, they may behave differently.
 
 .. seealso::
 
-   `winpexpect <https://pypi.python.org/pypi/winpexpect>`__ and `wexpect <https://gist.github.com/anthonyeden/8488763>`__
-     Two unmaintained pexpect-like modules for Windows, which work with a
-     hidden console.
+   `wexpect <https://pypi.org/project/wexpect/>`__ is a maintained
+   alternative for Windows. It supports pyton3.x
+   
+   `winpexpect <https://pypi.python.org/pypi/winpexpect>`__ is unmaintained
+   pexpect-like module, which work with a hidden console.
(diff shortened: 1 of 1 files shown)


doc/overview.rst
```diff
@@ -264,6 +264,8 @@ When run by ``PopenSpawn``, they may behave differently.
 
 .. seealso::
 
-   `winpexpect <https://pypi.python.org/pypi/winpexpect>`__ and `wexpect <https://gist.github.com/anthonyeden/8488763>`__
-     Two unmaintained pexpect-like modules for Windows, which work with a
-     hidden console.
+   `wexpect <https://pypi.org/project/wexpect/>`__ is a maintained
+   alternative for Windows. It supports pyton3.x
```
takluyver: Glad to hear it, but I don't think we need to explicitly mention Python 3 support in 2020 - I'd assume that any maintained module supports Python 3 unless it says otherwise.

doc/overview.rst
```diff
@@ -264,6 +264,8 @@ When run by ``PopenSpawn``, they may behave differently.
...
+   
+   `winpexpect <https://pypi.python.org/pypi/winpexpect>`__ is unmaintained
+   pexpect-like module, which work with a hidden console.
```
takluyver: work -> works (now that it's only describing one thing.

doc/overview.rst
raczben: However, this is the only Windows alternative, with Python 3 support, I let it go, but the link is essential. The old link of wexpect points to an unmaintained repo, while the new link shows the maintained package. (The maintained package's source is https://github.com/raczben/wexpect)

takluyver: Yep, I have no objection to updating the link.

takluyver:
I decided to remove the link to winpexpect entirely, now that you're maintaining wexpect. It's not helpful to point people to a project that stopped 9 years ago.

PR #629, opened 2020-03-12
Do not directly call python and use sys.executable

scarabeusiv:
This makes sure the tests and wrapper works on systems where there
is no python2 nor /usr/bin/python available


takluyver:
Seems reasonable, thanks.

Issue #631, opened 2020-03-28
REPLWrapTestCase.test_bash_env fails if bashrc does not export PS1


whitley-delamora:
The test `REPLWrapTestCase.test_bash_env` in `tests/test_replwrap.py` fails if the users bashrc does not have an environment variable `PS1`.

The variable `PS1` is not a environment variable per se and thus not shown by `env` (cf. [this stackoverflow question](https://stackoverflow.com/questions/41931855/why-env-does-not-print-ps1-variable)).

It can be made into an environment variable by exporting it with `export PS1`, which is some have in their bashrc but others (like me) don't.

To make the test consistent either

* test for a real environment variable, or
* export it before running `env` in the test, thus not relying on the user configuration.

Something like:
```
    def test_bash_env(self):
        """env, which displays PS1=..., should not mess up finding the prompt.
        """
        bash = replwrap.bash()
        res = bash.run_command("export PS1")  # <--- This line is new
        res = bash.run_command("env")
        self.assertIn('PS1', res)
        res = bash.run_command("echo $HOME")
        assert res.startswith('/'), res
```
should suffice.

takluyver:
Thanks @whitley-delamora . This makes sense, and your proposed fix for the test looks good. Do you want to make a pull request with this change?

whitley-delamora:
Wow, what a quick response.

> Do you want to make a pull request with this change?

Been there, done that. Thx.

takluyver:
Thanks. I should say I'm not usually that quick, I just happened to have time to read and respond to your message. :wink:

PR #632, opened 2020-03-28
REPLWrapTestCase.test_bash_env(): export variable PS1

whitley-delamora:
The test test_bash_env() relies on PS1 being an environment variable.
Therefore, `export PS1` making it always an environment variable.
Before this, the test relied on the users bashrc to do this.

The variable PS1 is (by default) not an _environment_ variable and thus
not shown by `env`. Any variable can be made into an environment
variable by exporting it. Some people do this for PS1 in their bashrc.

Closes: #631


takluyver:
For some reason, Travis CI doesn't seem to have shown up on this PR, but it did run, and all the tests passed:
https://travis-ci.org/github/pexpect/pexpect/builds/668045493

PR #634, opened 2020-04-06
Format entire project with black

xmunoz:
- ~~Add black to CI~~, and document usage in DEVELOPMENT.rst.
- Clean up .travis.yml: remove deprecated `sudo` key and install testing requirements from requirements-testing.txt
  Ref: https://docs.travis-ci.com/user/reference/overview/#deprecated-virtualization-environments
- Add pypy3 to CI

Edit: I gave up on adding black to CI because of of it's installation dependency on >=python3.6. A documented policy is probably enough for now, and the cleaned up formatting is a still valuable from a development experience perspective. If there's interested in adding something to CI, `flake8` is compatible with all of the python versions listed in .travis.yml, and provides a similar kind of pass/fail output. That could be added in a separate PR.


xmunoz:
I appreciate that this CL, because of its size, may be hard to review. To verify that it does the things I purport in the description, you could try manually applying my changes to .travis.yml, DEVELOPERS.rst and testing-requirements.txt. Once black is installed, run `black .` from the project root. Then you can generate a patch using `git format-patch HEAD~`. If you remove the header, our patch files should be identical.

xmunoz:
I suspect coveralls is bad at math. A formatting change *shouldn't* decrease coverage, even by 0.01%.

takluyver:
I'll leave this open for now for in case the other maintainers feel differently, but my own leaning is a blanket 'no thanks' to massive reformattings of existing code. I'm fine with people tidying up code they're touching anyway, but reformatting everything makes things like `git blame` much less useful, and causes unnecessary merge conflicts.

I'm not concerned by the coverage change, but I expect it's caused by places where reformatting breaks up one line into several, or vice versa.

Red-M:
I understand the idea of black to just uni-formally convert any code into it's own style which gives the entire project a single code style however I don't agree with its choices on said style.

Personally I'm of the mind that people should be following the style of the code they're changing when submitting PRs. As much as that isn't particularly enforce-able for a project I see it as being polite to the project you're contributing to.

To summarise I agree with @takluyver in addition to his reasons as well.

xmunoz:
> I'm fine with people tidying up code they're touching anyway, but reformatting everything makes things like `git blame` much less useful, and causes unnecessary merge conflicts.

`git blame` [support revision hashes](https://git-scm.com/docs/git-blame#_specifying_ranges), the default is simply `HEAD`. You can also glean this information from `git log`. Something like this should do what you want:

    git log -L line_num_start,line_num_end:filename

As for the other concern, I took a look at the open PRs. All from the last 6 months are 1-line changes in existing files, adding new files. or modifying .travis.yml. Cleaning up merge conflicts for any of these cases should not be particularly taxing.

> I understand the idea of black to just uni-formally convert any code into it's own style which gives the entire project a single code style however I don't agree with its choices on said style.

Which style choices do you disagree with? Black [is configurable](https://black.readthedocs.io/en/stable/pyproject_toml.html#configuration-format) to a certain extent. 

> Personally I'm of the mind that people should be following the style of the code they're changing when submitting PRs. As much as that isn't particularly enforce-able for a project I see it as being polite to the project you're contributing to.

What is the style of this source code? At a glance, it's extremely inconsistent with line lengths , comment syntax, and white space. I also couldn't find the style of the code documented anywhere.

xmunoz:
Also, something like code style _can_ and __should__ be enforced. That is was I was hoping to achieve with this PR and subsequent PRs.

PR #644, opened 2020-05-09
Add history URL to project metadata

scop:
Background info at
  https://github.com/pypa/warehouse/pull/7882#issue-412444446


takluyver:
Thanks!

takluyver:
I belatedly noticed that `history` wasn't one of the options you added in your Warehouse PR. Do you want to make another PR here to change the name to `changelog` or something?

scop:
I have another warehouse PR pending for that: https://github.com/pypa/warehouse/pull/7896

In any case, I think the link should match the title of the thing it links to, so in that sense I think it's fine as is. But personally I'd use "Changelog" or "Change log" (IIRC the former is more common) in both the link and the doc if I was starting a project, not sure if it's worth tweaking once it's out like it is in this case (there's also the history.html filename to consider, so... meh).

scop:
Note BTW that currently `History` as the title will work just fine too, warehouse will just show the readthedocs icon for it until the mentioned PR is merged.

takluyver:
OK, thanks.

PR #650, opened 2020-06-19
PR: Allow PopenSpawn to redirect stderr to a file

andfoy:
This PR allows to access the value of stderr, which may be useful to extract the logs of processes that write debug information to stderr

cc @ccordoba12


Red-M:
Please add at least one test for this new feature.

PR #663, opened 2020-10-19
fix `test_pager_as_cat` test case on Solaris

kulikjak:
On Solaris (and possibly other systems as I don't think this is standardized), the `sleep` manual page uses a lowercase name in the header, hence `SLEEP` is not found.

After this change, the test case will look for the section name, which AFAICT is always uppercase.

Alternatively, the `assert` can be changed to something like this:
`assert 'SLEEP' in res.upper(), res`

EDIT: The section name approach doesn't work (not sure why it seemed to work yesterday), so I changed it to the `res.upper()` approach instead. Thinking about that now, it is more robust and platform independent as well.

### DIFF
--- a/tests/test_replwrap.py
+++ b/tests/test_replwrap.py
@@ -39,7 +39,7 @@ class REPLWrapTestCase(unittest.TestCase):
         " PAGER is set to cat, to prevent timeout in ``man sleep``. "
         bash = replwrap.bash()
         res = bash.run_command('man sleep', timeout=5)
-        assert 'NAME' in res, res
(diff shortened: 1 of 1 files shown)


rantala:
Also the test should check that `man` is available:
`FAILED tests/test_replwrap.py::REPLWrapTestCase::test_pager_as_cat - AssertionError: bash: man: command not found`

kulikjak:
You are right, but I guess that a separate PR/Issue should be opened for that as that is not a problem on Solaris and fixes a slightly different thing.

Red-M:
Thanks!

PR #684, opened 2021-04-17
Set daemon attribute instead of using setDaemon method that was deprecated in Python 3.10

### LINKED ISSUE #683: threading.Thread.setDaemon has been deprecated in favor of setting daemon attribute directly in Python 3.10
Ref : python/cpython#25174

https://github.com/pexpect/pexpect/blob/02c38338b4ade8b0b0e85be613bc2b9dbfe68a7b/pexpect/popen_spawn.py#L60

tirkarthi:
Fixes #683


Red-M:
Thanks!

PR #690, opened 2021-06-19
switch to get_running_loop from get_event_loop to clear deprecation warning in python 3.10

### LINKED ISSUE #686: Python 3.10 : Deprecation warning regarding no current event loop
Ref : https://github.com/python/cpython/pull/25918

```
tests/test_async.py: 18 warnings
  /root/checked_repos_clone/pexpect/tests/test_async.py:15: DeprecationWarning: There is no current event loop
    return asyncio.get_event_loop().run_until_complete(coro)
```

Narendra-Neerukonda:
Fixes #686

### DIFF
--- a/tests/test_async.py
+++ b/tests/test_async.py
@@ -12,7 +12,7 @@ from pexpect import replwrap
 from .PexpectTestCase import PexpectTestCase
 
 def run(coro):
-    return asyncio.get_event_loop().run_until_complete(coro)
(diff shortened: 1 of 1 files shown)


Red-M:
Can a python version check be added as [`asyncio.get_running_loop()`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_running_loop) is new in python 3.7 and will break older versions of python. This is simply a convenience for older versions of python.

PR #691, opened 2021-06-30
pxssh.login: allow using implicit SSH configuration

rmelotte:
With OpenSSH clients, configuration data can come from multiple
sources (see man ssh_config):
1.   command-line options
2.   user's configuration file (~/.ssh/config)
3.   system-wide configuration file (/etc/ssh/ssh_config)

This means for example that the configuration for a particular host
could come from a user-defined configuration file, while for another
host it comes from the system-wide configuration file.

Besides, pexpect forces the use of either a username, or a specific
configuration file that will be parsed and checked for the host (among
other things).

As a consequence, it's currently not possible to use pexpect like a
regular OpenSSH client would be used, with multiple hosts potentially
defined in different configuration files.

To address this issue, instead of forcing the use of either a username
or a configuration file, just let the ssh client handle it. Also
remove the associated test.

This also allows the configuration files that are implicitly used to
contain more complex directives that are not currently supported when
parsing `ssh_config` (e.g. the `Include` directive).

### DIFF
--- a/pexpect/pxssh.py
+++ b/pexpect/pxssh.py
@@ -360,9 +360,8 @@ class pxssh (spawn):
 
         if username is not None:
             ssh_options = ssh_options + ' -l ' + username
-        elif ssh_config is None:
-            raise TypeError('login() needs either a username or an ssh_config')
-        else:  # make sure ssh_config has an entry for the server with a username
+        # If ssh_config was provided, make sure it has an entry for the server with a username.
+        elif ssh_config is not None:
             with open(ssh_config, 'rt') as f:
                 lines = [l.strip() for l in f.readlines()]
 
--- a/tests/test_pxssh.py
+++ b/tests/test_pxssh.py
@@ -99,16 +99,6 @@ class PxsshTestCase(SSHTestBase):
         if not '-F '+config_path in string:
             assert False, 'String generated from SSH config passing is incorrect.'
 
-    def test_username_or_ssh_config(self):
-        try:
-            ssh = pxssh.pxssh(debug_command_string=True)
-            temp_file = tempfile.NamedTemporaryFile()
-            config_path = temp_file.name
-            string = ssh.login('server')
-            raise AssertionError('Should have failed due to missing username and missing ssh_config.')
-        except TypeError:
-            pass
-
     def test_ssh_config_user(self):
         ssh = pxssh.pxssh(debug_command_string=True)
         temp_file = tempfile.NamedTemporaryFile()


Red-M:
Implicit configuration isn't a good idea, particularly when automatic processes are involved that are handling secrets such as passwords.

If this is required, I'd suggest using `pexpect.pxssh.psxssh()`'s `options` argument instead to load ssh config files. If more than one file is required that isn't a system default (eg, `~/.ssh/config` and `/etc/ssh/ssh_config`, which should already be loaded by `ssh` anyhow), I'd suggest placing includes to allow such a structure.

rmelotte:
> If this is required, I'd suggest using pexpect.pxssh.psxssh()'s options argument instead to load ssh config files. If more than one file is required that isn't a system default (eg, ~/.ssh/config and /etc/ssh/ssh_config, which should already be loaded by ssh anyhow), I'd suggest placing includes to allow such a structure.

That's what this PR is trying to do: at the moment it's not possible to use system defaults (eg, `~/.ssh/config` and `/etc/ssh/ssh_config`) without a username because pxssh requires either a username, or an SSH config.
If an SSH config is provided, then system defaults cannot be used anymore.

In my case I'm trying to use system defaults only, without explicitly providing a username.

Note that with this PR it's still possible to use an explicit username or SSH config file, only the requirement to provide one or the other is removed.

Red-M:
Yes, the function you're asking for should be explicitly requested, not via providing neither.

Red-M:
Upon thinking about this further I think you should try turning on openSSH-client's verbose mode as I'm pretty sure that system defaults are always loaded from `/etc/ssh/ssh_config` and then `~/.ssh/config`, provided that the files are readable.

rmelotte:
Thanks for the update!

> as I'm pretty sure that system defaults are always loaded from /etc/ssh/ssh_config and then ~/.ssh/config, provided that the files are readable.

They are read, unless a specific per-user configuration file is provided.

From the man page:
>-F configfile
>             Specifies an alternative per-user configuration file.  If a configuration file is given on the command line, the system-wide configuration file (/etc/ssh/ssh_config) will be ignored.  The default for the per-user configuration file is ~/.ssh/config.

This makes it impossible to use pexpect like you could use OpenSSH (without manually providing a User or configuration file, and instead using the combination of the default configuration files).

As a consequence there are a number of use cases that are valid for OpenSSH, but not for pexpect.

Just to give an example:
- I might have only a Host defined in `~/.ssh/config` (with no `User`), with a generic rule that will define the username for me in `/etc/ssh/ssh_config`.
  Let's say `~/.ssh/config` is:
  ```
  Host home-laptop
    Hostname 192.168.1.2
  ```
  And `~/etc/ssh/ssh_config` is:
  ```
  Host home-*
    User homeuser
  ```
  OpenSSH has no problem with that, as the configuration it uses is the combination of `/etc/ssh/ssh_config` and `~/.ssh/config`. I just have to provide the Host (`home-laptop`), and OpenSSH will handle the rest (i.e. `ssh home-laptop` works, it connects to `home-laptop` with the username `homeuser`).
  With pexpect on the otherhand, I either have to provide a username or a configuration file:
    - If I provide a username, then there is no point in configuring it in `/etc/ssh/ssh_config`, and I'll have to duplicate the logic of mapping usernames to host in my python application (which is undesirable of course).
    - If I provide a configuration file, it either has to be `~/.ssh/config` or `/etc/ssh/ssh_config`.
      In the first case:
      ```
      import os.path
      import pexpect.pxssh

      with pexpect.pxssh.pxssh() as shell:
          shell.login("home-laptop", ssh_config=os.path.expanduser("~/.ssh/config"))
      ```
      pexpect will refuse to connect to the host, because the configuration file doesn't provide a username:
      ```
      TypeError: login() ssh_config has no user entry for home-laptop
      ```
      Even if we remove the check for `User` when parsing the configuration file, pexpect would use the incorrect username (because of the `-F` option, OpenSSH will not read `/etc/ssh/ssh_config` so the current local username will be used instead).
      In the second case:
      ```
      import pexpect.pxssh

      with pexpect.pxssh.pxssh() as shell:
          shell.login("home-laptop", ssh_config="/etc/ssh/ssh_config")
      ```
      The Host `home-laptop` is not known to pexpect (and OpenSSH) because `~/.ssh/ssh_config` is not read.
      So it will fail with:
      ```
      TypeError: login() ssh_config has no Host entry for home-laptop
      ```
This is why I propose to not enforce the use of either a username or a specific ssh_config, and let OpenSSH handle it (so that if something works with the local ssh client, it also "just works" with pexpect).
If you would prefer to have an optional parameter for it I can try to add it (even though IMO it should be the default behavior).

As a side note (but it's again just my opinion and it's not for this PR) I think pexpect would better not parse the SSH configuration file at all, as the parsing logic is always going to be a subset of what OpenSSH does. For example what if `Include` is used to include other configuration files? pexpect won't read the included files, so it also won't work in cases where the OpenSSH client will.

PR #698, opened 2021-09-22
Make test_expect.py work on POSIX systems that are not Linux based

stesser:
Linux specific commands ans options are used in the test_before_across_chunks function (/bin/bash; nl with long options) which make this test fail on generic POSIX systems. No bash specific features or GNU nl features are used, therefore it is easily possible to make the test run on any POSIX system.

There was an assumption that STDIO buffers are large enough to hold some 10 KB of data. I assume that these buffers are 16 KB on Linux by default, but this is not true for other POSIX systems. The default STDIO buffer size is 4 KB on FreeBSD, for example, which leads to "cat" being blocked waiting to drain its output buffer before it accepts any further input.

The test does not depend on any particular amount of data or characters per line, therefore lines of 39 text characters plus line end characters should be sufficient and does allow systems with 4 KB default STDIO buffer size to successfully execute this test.

### DIFF
--- a/pexpect/bashrc.sh
+++ b/pexpect/bashrc.sh
@@ -14,3 +14,5 @@ PS1="$"
 
 # Unset PROMPT_COMMAND, so that it can't change PS1 to something unexpected.
 unset PROMPT_COMMAND
+
+bind 'set enable-bracketed-paste off'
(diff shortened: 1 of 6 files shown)


Red-M:
I'm not seeing the results of the tests from travis/coveralls. Have you run these tests locally?

Red-M:
Also noting that there might be more work in terms of testing targets (adding them to the existing CI jobs) to ensure that these changes do no regress.

Red-M:
The GHA won't run unless the PR is re-opened.

Red-M:
Please address any issues that arise from the tests.
Please also note that if I don't hear anything back in the next year, I'll be closing this.

PR #710, opened 2022-02-05
async for waitnoecho/send/sendline, method naming convention for _async.py

andrews239:
1) method naming convention for _async.py

ClassName_MethodName_async
```
class spawn(SpawnBase):
    def waitnoecho(self, timeout=-1, async_=False):
        if async_:
            from ._async import spawn__waitnoecho_async
            return spawn__waitnoecho_async(self, timeout, end_time)
        else:
            return self._waitnoecho(timeout, end_time)
```
(ClassName is spawn, MethodName is _waitnoecho)

2) async version for:
     spawn.waitnoecho()
     spawn.send()
     spawn.sendline()

3) remove time.sleep() from async methods

4) Results of testing.
20 spawn and 50 fast commands per spawn.
asycsend = True     --     5 sec (theory >2.5 sec)
asycsend = False    --   50 sec (theory >50 sec)

```
import asyncio
import pexpect
import time

async def test_send(num, amode):
    p = pexpect.spawn('python', encoding='utf8')
#    p.delaybeforesend = None
    for j in range(1,50):
        await p.expect(r'>>>', async_=True)
        if amode:
            await p.sendline('print("foo bar")', async_=True)
        else:
            p.sendline('print("foo bar")')
    await p.expect(r'>>>', async_=True)
    await p.sendline('exit()', async_=True)
    await p.expect(pexpect.EOF, async_=True)

async def test_loop(amode=False):
    start = time.time()
    tasks = []
    for i in range(1, 20):
        tasks.append(test_send(i, amode))
    await asyncio.wait(tasks)
    print('All Tasks done in {} seconds (asyncsend={})\n'.format(time.time() - start, amode))

loop = asyncio.get_event_loop()
loop.run_until_complete(test_loop())
loop.run_until_complete(test_loop(True))
```

```
All Tasks done in 49.074270486831665 seconds (asyncsend=False)
All Tasks done in 4.687046527862549 seconds (asyncsend=True)
```

### DIFF
--- a/pexpect/_async.py
+++ b/pexpect/_async.py
@@ -1,11 +1,12 @@
 import asyncio
 import errno
 import signal
+import os
 
 from pexpect import EOF
 
 @asyncio.coroutine
-def expect_async(expecter, timeout=None):
+def Expecter_expect_loop_async(expecter, timeout=None):
     # First process data that was previously read - if it maches, we don't need
     # async stuff.
     idx = expecter.existing_data()
@@ -28,13 +29,13 @@ def expect_async(expecter, timeout=None):
         return expecter.timeout(e)
 
 @asyncio.coroutine
-def repl_run_command_async(repl, cmdlines, timeout=-1):
+def REPLWrapper__run_command_async(repl, command, cmdlines, timeout=-1):
     res = []
-    repl.child.sendline(cmdlines[0])
+    yield from repl.child.sendline(cmdlines[0], async_=True)
     for line in cmdlines[1:]:
         yield from repl._expect_prompt(timeout=timeout, async_=True)
         res.append(repl.child.before)
-        repl.child.sendline(line)
+        yield from repl.child.sendline(line, async_=True)
 
     # Command was fully submitted, now wait for the next prompt
     prompt_idx = yield from repl._expect_prompt(timeout=timeout, async_=True)
@@ -42,7 +43,8 @@ def repl_run_command_async(repl, cmdlines, timeout=-1):
         # We got the continuation prompt - command was incomplete
         repl.child.kill(signal.SIGINT)
         yield from repl._expect_prompt(timeout=1, async_=True)
-        raise ValueError("Continuation prompt found - input was incomplete:")
+        raise ValueError("Continuation prompt found - input was incomplete\n"
+                             + command)
     return u''.join(res + [repl.child.before])
 
 class PatternWaiter(asyncio.Protocol):
@@ -101,3 +103,25 @@ class PatternWaiter(asyncio.Protocol):
             self.eof_received()
         elif exc is not None:
             self.error(exc)
+
+@asyncio.coroutine
+def spawn__waitnoecho_async(spawn, timeout, end_time):
+    while True:
+        if not spawn.getecho():
+            return True
+        if timeout < 0 and timeout is not None:
+            return False
+        if timeout is not None:
+            timeout = end_time - time.time()
+        yield from asyncio.sleep(0.1)
+
+@asyncio.coroutine
+def spawn__send_async(spawn, s):
+    if spawn.delaybeforesend is not None:
+        yield from asyncio.sleep(spawn.delaybeforesend)
+
+    s = spawn._coerce_send_string(s)
+    spawn._log(s, 'send')
+
+    b = spawn._encoder.encode(s, final=False)
+    return os.write(spawn.child_fd, b)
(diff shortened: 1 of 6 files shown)


Red-M:
While this PR is good overall, please don't change internal class names like that.

pexpect/_async.py
```diff
@@ -1,11 +1,12 @@
 import asyncio
 import errno
 import signal
+import os
 
 from pexpect import EOF
 
 @asyncio.coroutine
-def expect_async(expecter, timeout=None):
+def Expecter_expect_loop_async(expecter, timeout=None):
```
Red-M, line 9: Unless there is a technical reason to change the names of these functions, please revert these.

pexpect/expect.py
```diff
@@ -40,7 +40,7 @@ def do_search(self, window, freshlen):
                 spawn._buffer.write(window[-maintain:])
 
     def existing_data(self):
-        # First call from a new call to expect_loop or expect_async.
+        # First call from a new call to expect_loop or Expecter_expect_loop_async.
```
Red-M, line 43: Please see previous comment about function name changes.

pexpect/replwrap.py
```diff
@@ -89,9 +89,12 @@ def run_command(self, command, timeout=-1, async_=False):
             raise ValueError("No command was given")
 
         if async_:
-            from ._async import repl_run_command_async
-            return repl_run_command_async(self, cmdlines, timeout)
+            from ._async import REPLWrapper__run_command_async
```
Red-M, line 92: Please see previous comment about function name changes.

pexpect/spawnbase.py
```diff
@@ -366,8 +366,8 @@ def expect_list(self, pattern_list, timeout=-1, searchwindowsize=-1,
 
         exp = Expecter(self, searcher_re(pattern_list), searchwindowsize)
         if async_:
-            from ._async import expect_async
-            return expect_async(exp, timeout)
+            from ._async import Expecter_expect_loop_async
```
Red-M, line 369: Please see previous comment about function name changes.

pexpect/spawnbase.py
```diff
@@ -415,8 +415,8 @@ def prepare_pattern(pattern):
 
         exp = Expecter(self, searcher_string(pattern_list), searchwindowsize)
         if async_:
-            from ._async import expect_async
-            return expect_async(exp, timeout)
+            from ._async import Expecter_expect_loop_async
```
Red-M, line 418: Please see previous comment about function name changes.

pexpect/_async.py
andrews239: Hello, Red-M.

Thanks for the quick response. 

See my first point...

I need to add some methods to _async.py.
For choosing names, I need a naming convention...
The current convention looks like "random name".

I have several options:
1) leave PR as is...
2) return original method names, and for new methods use random names like "def _lkjdsfldskjds()"
3) use a separate file instead of _async.py

Isn't naming a sufficient technical reason?

Notice how easy it is to compare the synchronous and asynchronous version 
of a method using the schema I suggested:

```
--- REPLWrapper._run_command    2022-02-06 13:00:58.640969771 -0800
+++ REPLWrapper__run_command_async      2022-02-06 13:01:30.472217769 -0800
@@ -1,17 +1,18 @@
-class REPLWrapper(object):
-    def _run_command(self, command, cmdlines, timeout):
+@asyncio.coroutine
+def REPLWrapper__run_command_async(repl, command, cmdlines, timeout=-1):
         res = []
-        self.child.sendline(cmdlines[0])
+    yield from repl.child.sendline(cmdlines[0], async_=True)
         for line in cmdlines[1:]:
-            self._expect_prompt(timeout=timeout)
-            res.append(self.child.before)
-            self.child.sendline(line)
+        yield from repl._expect_prompt(timeout=timeout, async_=True)
+        res.append(repl.child.before)
+        yield from repl.child.sendline(line, async_=True)

         # Command was fully submitted, now wait for the next prompt
-        if self._expect_prompt(timeout=timeout) == 1:
+    prompt_idx = yield from repl._expect_prompt(timeout=timeout, async_=True)
+    if prompt_idx == 1:
             # We got the continuation prompt - command was incomplete
-            self.child.kill(signal.SIGINT)
-            self._expect_prompt(timeout=1)
-            raise ValueError("Continuation prompt found - input was incomplete:\n"
+        repl.child.kill(signal.SIGINT)
+        yield from repl._expect_prompt(timeout=1, async_=True)
+        raise ValueError("Continuation prompt found - input was incomplete\n"
                              + command)
-        return u''.join(res + [self.child.before])
+    return u''.join(res + [repl.child.before])

```

PR #713, opened 2022-02-12
Disallow to use "with" clause with PopenSpawn class

dolfinus:
#695 was closed because `Popen.Spawn` implementation could cause blocking the current thread.

So I've removed `__enter__` and `__exit__` methods from this class to disallow users to use `PopenSpawn` as context manager.


dolfinus:
@Red-M Could you check, please?

Red-M:
This PR does not do what it describes as it adds code to other classes while removing the context manager from spawnbase.
If you wish to have this, you're more than welcome to have a wrapper class to add this in your own code as this is what this meant to support (`pexpect.spawn()` does this already as an example).

PR #722, opened 2022-08-23
Ensure test_run.py works when system-wide bashrc produces output

waveform80:
The `--rcfile` option to bash does not prevent the system-wide `/etc/bash.bashrc` script from running (contrary to the Debian man-page). This causes `test_run.py` to fail when the system-wide script produces output (as is the case in an Ubuntu image where the regular user belongs to the "sudo" group and the system-wide script produces an informational message about `sudo` usage).

Rather than rely on `--rcfile`, this patch changes the test to use `--norc` (which does suppress the execution of the system-wide bashrc), and changes `PS1` via the inherited environment instead (this also slightly simplifies the fixture setup too as `tempfile` is no longer required).

### DIFF
--- a/tests/test_run.py
+++ b/tests/test_run.py
@@ -22,7 +22,6 @@ PEXPECT LICENSE
 import pexpect
 import unittest
 import subprocess
-import tempfile
 import sys
 import os
 from . import PexpectTestCase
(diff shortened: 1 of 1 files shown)


Red-M:
Apologies, re-openning is the only way to get github to realise it needs to run the test suite.

PR #726, opened 2022-10-22
Add a serial spawn interface

antoniovazquezblanco:
First contribution to this project. I hope this code is of interest.

Please, let me know if there is interest on this PR and any improvement I can make to get this merged.

Thanks.

### DIFF
--- a/pexpect/serial_spawn.py
+++ b/pexpect/serial_spawn.py
@@ -1,0 +1,95 @@
+import os
+import sys
+import time
+import threading
+from pexpect.fdpexpect import SpawnBase
+from pexpect.exceptions import EOF, TIMEOUT, ExceptionPexpect
+
+PY3 = (sys.version_info[0] >= 3)
+
+
+class SerialSpawn(SpawnBase):
+    '''This is a serial interface class interface for Pexpect. Use this class to
+    control an existing python serial connection. '''
+    def __init__(self, ser, args=None, **kwargs):
+        if not ser.is_open:
+            raise ExceptionPexpect('serial port is not ready')
+        super().__init__(None, **kwargs)
+        self.ser = ser
+        self.name = f'<serial port {ser.port}>'
+        self.closed = False
+        self.interacting = threading.Event()
+
+    def close(self):
+        if not self.ser.is_open:
+            return
+        self.flush()
+        self.ser.close()
+        self.closed = True
+
+    def isalive(self):
+        return self.ser.is_open
+
+    def send(self, s):
+        s = self._coerce_send_string(s)
+        self._log(s, 'send')
+        b = self._encoder.encode(s, final=False)
+        return self.ser.write(b)
+    
+    def read_nonblocking(self, size=1, timeout=-1):
+        if not self.isalive():
+            raise EOF('End Of File (EOF).')
+        if timeout and timeout < 0:
+            timeout = self.timeout
+        if timeout and timeout < 0:
+            timeout = 0
+        self.ser.timeout = timeout
+        s = self.ser.read(n)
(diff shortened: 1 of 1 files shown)


Red-M:
You may want to take a look at #745 as I'm not going to accept this without unit tests.

antoniovazquezblanco:
Sorry for the long halt. I am coming back to this one...
The interface i wrote expects a pyserial object. I did not provide any typing information. For this reason, there is "no declared pyserial dependency" but the module won't work without it...
I have some questions for you:
1. Do you want me to add an "optional" pyserial dep to the project?
2. If so, do you have any preference on how this interface should probe for pyserial availability?
3. Testing will require serial mockup. I was thinking on creating a pty (http://allican.be/blog/2017/01/15/python-dummy-serial-port.html) but this will not work for testing on windows. Are you ok with this or do you prefer a more complex solution that works on Windows?

Thanks

pexpect/serial_spawn.py
```diff
@@ -0,0 +1,99 @@
...
+        b = self._encoder.encode(s, final=False)
+        return self.ser.write(b)
+    
+    def read_nonblocking(self, size=1, timeout=-1):
+        if not self.isalive():
+            raise EOF('End Of File (EOF).')
+        if not timeout:
+            timeout = -1
+        limit_time = time.time() + timeout
+        while self.ser.in_waiting <= 0  and time.time() < limit_time:
```
Hoohaha, line 45: why not set serial.timeout before serial.read()?

antoniovazquezblanco: Nice suggestion. Implemented.

pexpect/serial_spawn.py
```diff
@@ -0,0 +1,93 @@
...
+        b = self._encoder.encode(s, final=False)
+        return self.ser.write(b)
+    
+    def read_nonblocking(self, size=1, timeout=-1):
+        if not self.isalive():
+            raise EOF('End Of File (EOF).')
+        if timeout and timeout < 0:
+            timeout = self.timeout
+        self.ser.timeout = timeout
+        s = self.ser.read(n)
```
Hoohaha, line 45: [pyserial timeout](https://pyserial.readthedocs.io/en/latest/pyserial_api.html)

Possible values for the parameter timeout which controls the behavior of [read()](https://pyserial.readthedocs.io/en/latest/pyserial_api.html#serial.Serial.read):

timeout = None: wait forever / until requested number of bytes are received
timeout = 0: non-blocking mode, return immediately in any case, returning zero or more, up to the requested number of bytes
timeout = x: set timeout to x seconds (float allowed) returns immediately when the requested number of bytes are available, otherwise wait until the timeout expires and return all bytes that were received until then.


Not sure what will happen if timeout=-1.
please look at the example usage of read() in code: https://github.com/pyserial/pyserial/blob/31fa4807d73ed4eb9891a88a15817b439c4eea2d/serial/threaded/__init__.py#L198

data = self.serial.read(self.serial.in_waiting or 1)

antoniovazquezblanco: Avoid the undefined behaviour. Do not wait in case of a negative value for timeout.

antoniovazquezblanco:
Ping @Red-M. Could you please answer my doubts in the comment above? Thanks :)

Red-M:
I think this might be better off as its own library for interacting with serial devices.

antoniovazquezblanco:
Given the last comment, it seems this does not suit the project. I have uploaded the files to https://github.com/antoniovazquezblanco/pexpect-serialspawn where I will begin to work on install files and infraestructure to distribute this independently.

PR #715, opened 2022-03-24
Convert @asyncio.coroutine to async def

hroncok:
This is required for Python 3.11+ support.

Fixes https://github.com/pexpect/pexpect/issues/677

PR #732, opened 2022-11-23
refactor(_async.py): introduce modern coroutine syntax

karrukola:
Python 3.5 introduced the await and async def keywords. This change imports the coroutines from two different places, depending on the running Python version.

This address comments previously put in pull requests:

- #715 
- #690 
-

### DIFF
--- a/tests/test_async.py
+++ b/tests/test_async.py
@@ -4,83 +4,90 @@ except ImportError:
     asyncio = None
 
 import gc
-import sys
 import unittest
+from sys import version_info as py_version_info
 
 import pexpect
 from pexpect import replwrap
+
 from .PexpectTestCase import PexpectTestCase
 
+
 def run(coro):
-    return asyncio.get_event_loop().run_until_complete(coro)
+    if py_version_info >= (3, 7):
+        # get_running_loop, new in 3.7, is preferred to get_event_loop
+        return asyncio.get_running_loop().run_until_complete(coro)
+    else:
+        return asyncio.get_event_loop().run_until_complete(coro)
+
 
 @unittest.skipIf(asyncio is None, "Requires asyncio")
 class AsyncTests(PexpectTestCase):
     def test_simple_expect(self):
-        p = pexpect.spawn('cat')
-        p.sendline('Hello asyncio')
-        coro = p.expect(['Hello', pexpect.EOF] , async_=True)
+        p = pexpect.spawn("cat")
+        p.sendline("Hello asyncio")
+        coro = p.expect(["Hello", pexpect.EOF], async_=True)
         assert run(coro) == 0
-        print('Done')
+        print("Done")
 
     def test_timeout(self):
-        p = pexpect.spawn('cat')
-        coro = p.expect('foo', timeout=1, async_=True)
+        p = pexpect.spawn("cat")
+        coro = p.expect("foo", timeout=1, async_=True)
         with self.assertRaises(pexpect.TIMEOUT):
             run(coro)
 
-        p = pexpect.spawn('cat')
-        coro = p.expect(['foo', pexpect.TIMEOUT], timeout=1, async_=True)
+        p = pexpect.spawn("cat")
+        coro = p.expect(["foo", pexpect.TIMEOUT], timeout=1, async_=True)
         assert run(coro) == 1
 
     def test_eof(self):
-        p = pexpect.spawn('cat')
-        p.sendline('Hi')
+        p = pexpect.spawn("cat")
+        p.sendline("Hi")
         coro = p.expect(pexpect.EOF, async_=True)
         p.sendeof()
         assert run(coro) == 0
 
-        p = pexpect.spawn('cat')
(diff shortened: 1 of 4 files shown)


piraz:
Nice.

Red-M:
Sorry for the spam, its the only way to restart the tests.

tests/test_async.py
```diff
@@ -4,83 +4,90 @@
...
+        # get_running_loop, new in 3.7, is preferred to get_event_loop
+        return asyncio.get_running_loop().run_until_complete(coro)
+    else:
+        return asyncio.get_event_loop().run_until_complete(coro)
+
 
 @unittest.skipIf(asyncio is None, "Requires asyncio")
 class AsyncTests(PexpectTestCase):
     def test_simple_expect(self):
-        p = pexpect.spawn('cat')
```
Red-M, line 20: Please do not include style changes like these, could you please revert these?

Red-M:
Could you address the issue(s) found in the tests?
I don't mind so much if the coverage drops by less than a percent as this is due to the nature of the expect loop but the functional tests should still pass.

tests/test_async.py
karrukola: Ack

karrukola:
Will do. at time of submission I could not get 100% pass from master, so I did not manage to get a clear view. Is there any specific Python version?

karrukola:
Hello there, I would need one more week, sorry.
Life got in between.

Red-M:
Don't worry, I wouldn't expect you or anyone else to force themselves to get things done.
Life happens and thats ok.
I don't expect myself to put (F)OSS work above my own personal life so no one here is going to pressure you to do anything.

Take your time, great software isn't made overnight :)

Red-M:
Still failing, it looks like the tests may need to start an event loop before they run?

karrukola:
> Still failing, it looks like the tests may need to start an event loop before they run?

Yes, that would require rework of the tests, though.
I understand this is because `asyncio.get_event_loop()` eventually creates one, while `get_running_loop()` that is the new recommended API, does not.
As a solution, right now, I reverted my original change on the test file. A deprecation warning would still be thrown, but it looks like they run on Python 3.11 on my PC.
On my local setup I have more failures than this CI pipelines, in the replwrap area.

Open questions about the rework of the tests:
1. should they be written using `unittest.IsolatedAsyncioTestCase`?
2. as you are running them with pytest, would you be open to the usage of `pytest-asyncio`? <- My preference would go for this because then the test execution would own the event loop, and the tests could be written as a series of `await`

Red-M:
I'd prefer `unittest.IsolatedAsyncioTestCase` simply because we have non-async tests as well and we need to run both in a single test run for coverage reasons.

I assume the replwrap errors on your local environment may be due to customisation but as long as there isn't actual issues and the GHA pipelines are fine, I think that is enough.

karrukola:
Hello @Red-M,
the CI failure on **pypy3.9** is in running `./tools/display-sighandlers.py`, so prior the execution of pytest.

As per **coverage** I understand you were OK to let it drop a bit with this change.

Let me know if you would like the unittests work (`unittest.IsolatedAsyncioTestCase`) to be done in this PR's scope; I would advise to address it in another PR.

Thanks

Red-M:
pypy tests have been broken for a number of years, I wouldn't pay it any mind now as it is unrelated to your changes.

The coverage change is due to the coverage only running on one version of python, so all of the pre-3.6 async code isn't tested which is a loss of coverage. I'll handle that via making the coverage test use multiple versions of python.

Happy for the tests to be changed in another PR, as long as the refactor makes sense.

Red-M:
@karrukola A gentle reminder that this won't be in a release until the unit tests are complete.

BoostCookie:
> @karrukola A gentle reminder that this won't be in a release until the unit tests are complete.

The only current error is with pypy. And the error (in tools/display-sighandlers.py) is that pypy defines some names in os.pathconf_names that give a ValueError in os.fpathconf(fd, name).
Is this kind of test even relevant? Would the current commit of pexpect actually break with pypy?

piraz:
It seems like @BoostCookie is on the latest pypy [v7.3.12](https://foss.heptapod.net/pypy/pypy/-/tags?sort=updated_desc&search=v7.3.12). Do we need to update any version and re-check this pull request?
I've been using this solution as a monkey patch, it would be great to have a release with this fix.
Do you need any help with that?

BoostCookie:
Yes, pypy has fixed the issue. To test it I've forked pexpect and retriggered the CI pipeline, and it now works for pypy.

The test for python 3.7 on other hand is now running for over 20 minutes :/
https://github.com/BoostCookie/pexpect/actions/runs/5516693915/jobs/10058349155

Red-M:
I don't know why you've latched onto this idea that its the pypy tests holding the release up.
https://github.com/pexpect/pexpect/pull/732#issuecomment-1466957747

This feature needs more tests to properly test the async API.

PR #733, opened 2022-11-28
Fix async keyword introduction release

andresdelfino:
I was fixing 3.7 support on a package, and didn't find this change at first because of the wrong release number.


Red-M:
I don't think that this is a good idea to change the history of the changelog.
A recent version's changelog (4.7) mentions async changes and I'd expect this to also be on the docs which it is and functions will mention if they have an async option.

PR #742, opened 2023-02-13
Fix SyntaxWarning in fakessh by using raw string

swt2c:


Red-M:
Sorry for the spam, its the only way to restart the tests.

PR #745, opened 2023-02-21
Socket Pexpect (for Windows support)

tapple:
This is exactly like fdpexpect, except it uses the cross-platform socket api. Thus it works on windows, unlike fdpexpect on socket.fileno()

Includes unit test and docs. I had to make a few unit test changes so the unit tests at least imported on windows. Most still fail, but they don't crash on import anymore


Red-M:
Please make sure that the tests pass, coverage changes that are less than 1% are ok but the more coverage the better.
At a cursory glance, this looks ok but I'm not sure how I feel about making the library partially windows compatible, it might be better that we refactor the library to be less platform dependant.

tapple:
I improved the coverage by making test_socket test both this and the existing fdspawn

Looks like 3.11 unit tests are blocked by !732; I can't do much about that

Red-M:
This looks good now, thank you for the PR.
One question I have is around what happens on windows when `import pexpect` is used?

tapple:
> This looks good now, thank you for the PR. One question I have is around what happens on windows when `import pexpect` is used?

I didn't change __init__.py at all. It didn't import much of anything before, and it still doesn't. pexpect.fdpexpect was never imported by "import pexpect" on any platform, and neither is pexpect.socket_pexpect. None of the local process manipulation code was imported by default on windows, and it still isn't:
```
from .exceptions import ExceptionPexpect, EOF, TIMEOUT
from .utils import split_command_line, which, is_executable_file
from .expect import Expecter, searcher_re, searcher_string

if sys.platform != 'win32':
    # On Unix, these are available at the top level for backwards compatibility
    from .pty_spawn import spawn, spawnu
    from .run import run, runu
```

PR #751, opened 2023-04-01
Add zsh convenience function to replwrap module

kislyuk:


Red-M:
Is the motivation for this to enable better Darwin support?

kislyuk:
Not specifically - we are using pexpect in shell integration test cases, for both bash and zsh. While zsh has become more popular because of Mac OS using it as the default, there are plenty of Linux-based zsh users too.

PR #753, opened 2023-04-08
Clean up temporary files after UnicodeTests

scfc:


Red-M:
Has this caused issues?

scfc:
Currently, the test (suite) leaves two temporary files behind after every run.  Is that behaviour intended?  If not, IMNSHO they should be cleaned up after the test has finished.

Red-M:
If you want this merged, please check the other tests as well.

scfc:
I don't know what I can do to resolve those "Expected — Waiting for status to be reported" checks.  My assumption would be that the repository is configured to use Travis CI, and that integration is no longer working.

PR #758, opened 2023-06-23
Update Docs Example Python 2 to Python 3

jasonandmonte:
Update example in doc to use Python 3 print statement. Also update the method calls to match the rest of the document.


Red-M:
Please do this to the whole documentation if you want this merged.

Issue #752, opened 2023-04-05
Async Test rework to handle warning from get_event_loop()


karrukola:
This is a spin-off of PR #732 
With Python 3.11 the current approach of calling on `get_event_loop()` throws a deprecation warning.
Moving to the recommended API `get_running_loop()` would make the tests fail as the older API would eventually create an event loop, while `get_running_loop()` does not.

Request by @Red-M is to rework the async tests to rely on `unittest.IsolatedAsyncioTestCase`

karrukola:
Apologies for the long radio silence; there have been a number of changes at work that made my focus go away.
I have asked @musthafak to provide the code within the next two weeks.

Red-M:
> Apologies for the long radio silence; there have been a number of changes at work that made my focus go away. I have asked @musthafak to provide the code within the next two weeks.

No worries, I do greatly appreciate the time and effort being provided here.

djuarezg:
Hi there! Do we have any update regarding this? I'd rather wait for this to be merged before manually patching things on the code.

Thank you for your help!

musthafak:
I've submitted a pull request at https://github.com/pexpect/pexpect/pull/764 to address and resolve this issue. This pull request also includes fixes for the pypy3.9 test failures.

Waiting for the review now.

PR #763, opened 2023-08-08
Add Python 3.5, 3.6 and 3.12.0-rc.1 to test matrix

### LINKED ISSUE #762: Support for new release
Hey @Red-M 

Is there something I can do to help having a new release out that works with 3.11 (and 3.12 - final is around the corner)?

fcollonval:
Closes #762

This adds:
- test against Python 3.5, 3.6 and 3.12.0-rc.1

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,11 +8,19 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-22.04
+    name: Test Python ${{ matrix.python-version }} on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


Red-M:
Yes, there was new features added in #732 and they need to have tests written for them

fcollonval:
Thanks @Red-M for the reply - I'm gonna need some help to be sure to get it right.

Looking at the [coverage](https://coveralls.io/github/fcollonval/pexpect), there is 8 lines missed in both `_async_*_await.py` that correspond to

image
> Coverage for Python 3.10 (but result is similar for other Python version).

Those lines are missing coverage prior to #732.

Could you provide pointers on what to do? There is the maintenance suggested in #752, is there something else?

PR #780, opened 2023-11-25
Add requirements-testing.txt to sdist

### LINKED ISSUE #768: sdist is missing requirements-testing.txt
The sdist package at PyPI is missing `requirements-testing.txt`.  Please add the missing file to sdist to make downstream testing easier.  Thank you.

mtelka:
Fixes #768.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,3 +4,4 @@ recursive-include examples *
 include .coveragerc README.rst LICENSE pexpect/bashrc.sh
 recursive-include tests *
 global-exclude __pycache__ *.pyc *~
+include requirements-testing.txt


Red-M:
This will also need to be added to [the package data files](https://github.com/pexpect/pexpect/blob/master/setup.py#L40).

mtelka:
> This will also need to be added to [the package data files](https://github.com/pexpect/pexpect/blob/master/setup.py#L40).

Why?  My understanding is that either `package_data` or `MANIFEST.in` should work, but we do not need both.

mtelka:
The https://setuptools.pypa.io/en/latest/userguide/datafiles.html#package-data says:

> Note that the data files specified using the `package_data` option neither require to be included within a [MANIFEST.in](https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html#using-manifest-in) file, nor require to be added by a revision control system plugin.

Red-M:
Fair enough, I've never really looked at MANIFEST.in but  `bashrc.sh`  is also there too.

PR #788, opened 2024-01-04
Add socket_pexpect read logging

edzius:
Using socket pexpect logs only `send*()` payload, however other
pexpect classes logs also received data. This change unifies logging
behaviour for socket pexpect.


nowayelse:
Nice! Another one forgotten action from SpawnBase.read_nonblocking is to decode when encoding is provided:
```
s = self._decoder.decode(s, final=False)
self._log(s, 'read')
```

PR #798, opened 2024-07-08
Permit import on Windows without Spawn error.

Xorba001:
Permit import on Windows without raising Spawn error :

from pexpect import ExceptionPexpect, TIMEOUT, EOF

if sys.platform != 'win32':
from pexpect import spawn
else:
class spawn:
def init(self, *args, **kwargs):
raise NotImplementedError("The 'spawn' function is not implemented on Windows.")

### DIFF
--- a/pexpect/pxssh.py
+++ b/pexpect/pxssh.py
@@ -20,7 +20,7 @@ PEXPECT LICENSE
 
 '''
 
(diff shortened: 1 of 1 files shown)


Xorba001:
I'm at a loss, cannot find a lint-fix makefile in the project. I just added a condition to the file.
This is mysterious for me. Please advise.

PR #831, opened 2026-06-18
fix: remove unused variables and PY3 compatibility stubs

decembercomposer697-hue:
7 fixes across 6 files:
- Remove unused variables (screen, mode, x)
- Replace PY3-compat code with direct Python 3 references
- Fix break outside loop in drain code

✅ 50 tests pass


wahajahmed010:
test

PR #838, opened 2026-09-19
update interact() to use str instead of bytes

mmgen:
This patch makes interact() usable again.

A quick and dirty fix. The devs will likely wish to improve on it.


mmgen:
@Red-M: Closed without a comment? Are you planning to fix this issue?

